PR #319, opened 2014-12-29
adapt to the interface change

greg0ire:
The addFilter() method signature has lost a default argument. Adapt.


dbu:
actually, as there is no typehint on the $type parameter, it can still be null so indeed this change has no effect.

lsmith77:
I guess we should release 1.2.1 based on the 1.2 branch ?

PR #321, opened 2015-01-18
Updated code for new TreeBrowserBundle

WouterJ:
This PR contains all changes required to use the new TreeBrowserBundle implementation: https://github.com/symfony-cmf/TreeBrowserBundle/pull/74

Currently, the upstream PR is in an early dev stage.

### DIFF
--- a/Resources/views/Form/form_admin_fields.html.twig
+++ b/Resources/views/Form/form_admin_fields.html.twig
@@ -146,38 +146,21 @@ file that was distributed with this source code.
 
 {% block doctrine_phpcr_odm_tree_widget %}
 {% include 'CmfTreeBrowserBundle:Base:tree.html.twig' %}
-<script type="text/javascript">
-    $(document).ready(function() {
-        var treeState = AdminTree.generateTreeStateArray("{{ value }}");
+<script>
+    jQuery(function ($) {
         var defaults = {% include 'SonataDoctrinePHPCRAdminBundle:Tree:routing_defaults.html.twig' with {'routing_defaults': routing_defaults} %};
 
-        SelectTree.initTree({
-            "selector": "#{{id}}-tree-selector",
-            "rootNode": "{{ root_node }}",
-            "ajax": {
-                "children_url": Routing.generate("_cmf_tree_{{ tree.alias }}_children", defaults)
-            },
-            "output": "#{{id}}-tree-selector-output",
-            "reset": "#{{id}}-tree-reset",
-            "path": {
-                "expanded":     treeState,
-                "preloaded":    treeState
-            },
-            "selected": '{{ value }}',
-            "selectRootNode": {{ select_root_node ? 'true' : 'false' }},
-            "routing_defaults": defaults
+        $('#{{ id }}-tree-selector').cmfTree({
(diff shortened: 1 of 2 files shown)


dbu:
lets bump the version requirement for the tree bundle to 2.0 to make sure we are aware this is a refactoring on that level.

WouterJ:
What's the plan for this PR: Merge it in master and continue to make seperate PRs for TreeBrowser 2.0 features or use this PR to push all TreeBrowser 2.0 features?

WouterJ:
Btw, could someone please rerun the tests for this PR?

dbu:
master is already at 2.0 so i guess we should merge this.

how much is missing on the tree browser to have a usable 2.0 release with sonata admin?

if we think it will take a long while until resources and tree have settled, we could just upgrade to sonata 2.4 as our phpcr-admin 2.0 release and do the tree refactoring in 2.1, to avoid having a long wait for a 2.0 release that supports base admin 2.4.

WouterJ:
@dbu TreeBrowser 2.0 could also work with the current tree backend in the SonataPHPCR admin bundle. We can then use Resource in 2.1.

What's missing from TreeBrowser is AdminTree support. If we decide to use the current tree backend instead of the Resource thing, I can start working on that at Monday.

dbu:
@dantleech what do you think? i would love to use the Resource stuff right away rather than having to re-refactor. but if this blocks a release for 6 months we have a problem.

dantleech:
@WouterJ what do you need from the ResourceRestBundle for the tree?

WouterJ:
@dantleech imo, there are 3 features missing features in the TreeBrowser bundle:

1. **AdminTree** (must have), this requires the knowledge of the CRUD functions in the front-end;
1. **Custom icons** (should have), this requires knowing the node type and optionally allowing custom payload properties like "icon_path";
1. **Path completion/pop-up** (could have), nothing more required here (started working on this yesterday).

As far as I know, the required functionality for (1) and (2) are not yet available in the ResourceRestBundle.

dantleech:
> AdminTree (must have), this requires the knowledge of the CRUD functions in the front-end;

This is my current priority, should be ready soon.

> Custom icons (should have), this requires knowing the node type and optionally allowing custom payload properties like "icon_path";

Firstly, I think this information belongs at a higher level. It is about mapping extra information to classes in gerneral - e.g. you should be able to retrieve the icon for a certain class in other contexts than the resource, this is why I started [this bundle](https://github.com/dantleech/ObjectInfoBundle). But I am unsure about that solution.

In the meantime, we already provide a payload "alias":

https://github.com/symfony-cmf/ResourceRestBundle/blob/master/Tests/Features/resource_api_phpcr_odm.feature#L35

Which could be used to infer an icon.

Otherwise we can add an extension to add such information.

WouterJ:
I just saw a commit from @lsmith77 where he updated the composer file to require TreeBrowser 2.0: https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/commit/5e37eb2a8889dee037b5597aac4d97d904c0a424 I think it's fine to merge this.

dbu:
thanks wouter!

PR #329, opened 2015-03-07
Use ResourceBundle as tree backend

WouterJ:
See https://github.com/symfony-cmf/TreeBrowserBundle/pull/84

### DIFF
--- a/Resources/views/Form/form_admin_fields.html.twig
+++ b/Resources/views/Form/form_admin_fields.html.twig
@@ -151,7 +151,16 @@ file that was distributed with this source code.
         var defaults = {% include 'SonataDoctrinePHPCRAdminBundle:Tree:routing_defaults.html.twig' with {'routing_defaults': routing_defaults} %};
 
         $('#{{ id }}-tree-selector').cmfTree({
-            data_url: Routing.generate('_cmf_tree_{{ tree.alias }}_children', defaults),
+            request: {
+                load: function (nodePath) {
+                    return {
+                        url: Routing.generate('_cmf_resource', jQuery.extend(defaults, {
+                            repositoryName: 'phpcrodm_repo',
+                            path: nodePath
+                        }))
+                    };
+                },
+            },
             path_output: '#{{ id }}-tree-selector-output'
         });
     });
--- a/Resources/views/Form/form_admin_fields.html.twig
+++ b/Resources/views/Form/form_admin_fields.html.twig
@@ -154,10 +154,10 @@ file that was distributed with this source code.
             request: {
                 load: function (nodePath) {
                     return {
-                        url: {{ path('_cmf_resource', {
+                        url: '{{ path('_cmf_resource', {
                             repositoryName: repository_name,
-                            path: '%path%'
-                        }) }}
+                            path: '__path__'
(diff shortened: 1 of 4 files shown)


Resources/views/Form/form_admin_fields.html.twig
```diff
@@ -151,7 +151,16 @@ file that was distributed with this source code.
         var defaults = {% include 'SonataDoctrinePHPCRAdminBundle:Tree:routing_defaults.html.twig' with {'routing_defaults': routing_defaults} %};
 
         $('#{{ id }}-tree-selector').cmfTree({
-            data_url: Routing.generate('_cmf_tree_{{ tree.alias }}_children', defaults),
+            request: {
+                load: function (nodePath) {
+                    return {
+                        url: Routing.generate('_cmf_resource', jQuery.extend(defaults, {
```
dantleech: I would really like to remove the fos js routing .. its needless overhead imo, we can generate the URL using ``path('_cmf_resource', array('repositoryName' => '%repo_name%', 'path' => '%path%'))`` and then use string substitution in the js.

Resources/views/Form/form_admin_fields.html.twig
```diff
@@ -151,7 +151,16 @@ file that was distributed with this source code.
...
+                            repositoryName: 'phpcrodm_repo',
```
dantleech: Guess this should be configurable

Resources/views/Form/form_admin_fields.html.twig
dantleech: Or maybe we should just remove the magic URL generation and just use query parameters, e.g. `http://foobar.com/resource/api?respository=my_repository&path=/foobar`

Resources/views/Form/form_admin_fields.html.twig
WouterJ: yes, indeed. Was thinking about that as well

WouterJ:
I've now changed the default repository name to `default`. With that, I think this PR is ready.

In future PRs, I'll try to move the tree form type to the TreeBrowserBundle, as that'll make more sense. This would also mean we can make the repository name configurable.

WouterJ:
Not sure why Scrutinizer fails, but this one seems ready imo.

dbu:
thanks a lot for this effort! do we migrate the sandbox to the new format as well? and does that mean we can attack all the TODOs where the admins show one level too much because the tree root node was never displayed?

WouterJ:
@dbu to be honest, I don't know about which TODOs you're talking (I kind of revisited the TreeBrowser code without looking too much into the current logic, but focusing more on what logic would be usefull for the admin). This PR opens ways to fix https://github.com/symfony-cmf/TreeBrowserBundle/issues/80 (admin tree), https://github.com/symfony-cmf/TreeBrowserBundle/issues/87 (move tree type to treeBrowser bundle) and https://github.com/symfony-cmf/TreeBrowserBundle/issues/75 (custom icons for specific types). These 3 are the issues that have kept the TreeBrowserBundle 2.0 from becoming close to stable. I should have lots of free time this week, so I'm hoping that I can work on them a lot the coming days.

PR #330, opened 2015-03-10
migrate ErrorElement

EmmanuelVella:
Related to https://github.com/sonata-project/SonataAdminBundle/pull/2792 (sonata-admin 2.4 branch)

@dbu @lsmith77 If this change is ok, I can submit others PRs for routing and block bundles as well.

### DIFF
--- a/Block/TreeBlockService.php
+++ b/Block/TreeBlockService.php
@@ -12,7 +12,6 @@
 namespace Sonata\DoctrinePHPCRAdminBundle\Block;
 
 use Sonata\AdminBundle\Form\FormMapper;
-use Sonata\AdminBundle\Validator\ErrorElement;
 use Sonata\BlockBundle\Block\BaseBlockService;
 use Sonata\BlockBundle\Block\BlockContextInterface;
 use Sonata\BlockBundle\Model\BlockInterface;
@@ -70,15 +69,5 @@ class TreeBlockService extends BaseBlockService
             'selected'         => null,
             'routing_defaults' => $this->defaults,
         ));
-    }
-
-    /**
-     * {@inheritDoc}
-     *
-     * NOOP as we do not edit and hence have nothing to validate.
-     */
-    public function validateBlock(ErrorElement $errorElement, BlockInterface $block)
-    {
-        // there is nothing to validate here
     }
 }


dbu:
@WouterJ this is fine to merge into master, right? master is already at sonata 2.4

@EmmanuelVella what do you want to change for routing and block bundle? it would really be nice if those can stay compatible with sonata 2.3 for a while... or are you not talking about the cmf block and routing bundle?

rande:
@dbu you can be compatible with Sonata 2.3 due to this BC break: https://github.com/sonata-project/SonataAdminBundle/commit/8486156f891ea49f48cd1979b6776014582480fa#diff-b6a3b9160f700d9778589e856c46433d

dbu:
can or can not? i have no worries about admin-phpcr being bound to specific admin. but the cmf routing bundle should not be locked to specific admin versions if ever possible.

EmmanuelVella:
@dbu It seems that the SonataCoreBundle has a ``Sonata\CoreBundle\Validator\ErrorElement`` class since the 2.3.1 version : https://github.com/sonata-project/SonataCoreBundle/commit/56a10e60f669b1d55bab504a76d92b5d20b0f8c1 So we can be compatible with a ``>=2.3.1`` requirement

In this case, maybe I should do this PR on the 1.2 branch instead ?

About the cmf routing and block bundle, I just propose to do the same thing (migrate from ``Sonata\AdminBundle\Validator\ErrorElement`` to ``Sonata\CoreBundle\Validator\ErrorElement``) if this PR is accepted.

dbu:
hm, the current phpcr admin is compatible with sonata 2.2 so this would mean we bump at least to 2.3... it looks to me like sonata core and sonata admin really should be called 3.0 if they change such essential method signatures. the BC way would be to keep the ErrorElement class in sonata admin but deprecate it and just extend the class from core bundle. and of course not change method signatures.

this is what we do with the phpcr-admin, as it needs to break BC in many ways. so this PR can be merged imo. for the cmf bundles, merging that change means we need to bump to the phpcr-admin 2.0. @WouterJ what is your take on that?

WouterJ:
Although PHPCR Admin 2.0 would be a lot nicer, I'm -1 for forcing all the CMF users to use SonataAdminBundle 2.4. As we all know, updating SonataAdminBundle 2.x has never been a very easy update.

It would have been really nice if 2.4 would be BC (or at least, easy to support 2.3 and 2.4). As forcing all CMF users to use 2.3 instead of 2.4 seems weird too.

Btw, can't we remove this method? Given that it's also empty in the BaseBlockService? https://github.com/sonata-project/SonataBlockBundle/blob/master/Block/BaseBlockService.php#L212-215

rande:
@dbu @lsmith77 give the go to include the new namespace, see https://github.com/sonata-project/SonataBlockBundle/pull/185 . Extending the ErrorElement class does not work as there is a mismatch with the class namespace even if the class extends the one from the core bundle. If you get it right, I will be happy to merge this

This has been done to remove hard dependency to the ``SonataAdminBundle`` which let the user with more flexibility and remove this kind of structural discussion.

EmmanuelVella:
@WouterJ Using the new namespace will work from admin 2.3.1, afaik there is not need to use the 2.4 branch.

However you are right, maybe we could simply remove these methods ? I just check, they all just throw an exception in the block and routing bundle (as well as the ``buildEditForm`` method)

EmmanuelVella:
Is it ok to finally remove these methods from this bundle and the routing/block one ?

dbu:
thanks emmanuel!

PR #337, opened 2015-04-24
Updated the PhpcrOdmTree to allow other compatible classes to be injecte...

wcluijt:
...d for the asset helper input

### DIFF
--- a/Tree/PhpcrOdmTree.php
+++ b/Tree/PhpcrOdmTree.php
@@ -15,6 +15,7 @@ use Doctrine\ODM\PHPCR\Document\Generic;
 use PHPCR\Util\NodeHelper;
 
 use PHPCR\Util\PathHelper;
+use Symfony\Bundle\FrameworkBundle\Templating\Helper\AssetsHelper;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException;
 use Symfony\Component\Translation\TranslatorInterface;
@@ -102,7 +103,7 @@ class PhpcrOdmTree implements TreeInterface
      * @param ModelManager $defaultModelManager to use with documents that have no manager
      * @param Pool $pool to get admin classes for documents from
      * @param TranslatorInterface $translator
-     * @param CoreAssetsHelper $assetHelper
+     * @param CoreAssetsHelper|AssetsHelper $assetHelper
      * @param array $validClasses list of the valid class names that may be
      *      used as tree "ref" fields
      * $param integer $depth depth to which grand children should be fetched,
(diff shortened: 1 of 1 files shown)


Tree/PhpcrOdmTree.php
```diff
@@ -114,7 +114,7 @@ public function __construct(
         ModelManager $defaultModelManager,
         Pool $pool,
         TranslatorInterface $translator,
-        CoreAssetsHelper $assetHelper,
```
dbu: can you please also adjust the phpdoc?

Tree/PhpcrOdmTree.php
```diff
@@ -127,6 +127,33 @@ public function __construct(
...
+
+        if (!$this->assetHelper instanceof CoreAssetsHelper) {
+            if (!(class_exists('\Symfony\Bundle\FrameworkBundle\Templating\Helper\AssetsHelper')
+                    && $this->assetHelper instanceof \Symfony\Bundle\FrameworkBundle\Templating\Helper\AssetsHelper
+                )
+                && !(class_exists('\Symfony\Component\Asset\Packages')
+                    && $this->assetHelper instanceof \Symfony\Component\Asset\Packages
+                )
+                && !(interface_exists('\Symfony\Component\Asset\PackageInterface')
+                    && $this->assetHelper instanceof \Symfony\Component\Asset\PackageInterface
```
dbu: afaik instanceof does not care whether the target class/interface exists. if i am not mistaken, you could simplify this code by just checking instanceof.

can you add a little bit of documentation explaining which symfony version needs what? its crazy that we have 4 different classes/interfaces that could be passed.

wcluijt:
I simplified the `instanceof` check.  I suppose at a later date the `Symfony\Component\Asset\Packages` and `Symfony\Component\Asset\PackageInterface` could be added to be compatible with the Symfony Asset component.

Tree/PhpcrOdmTree.php
```diff
@@ -127,6 +127,23 @@ public function __construct(
 
         $this->depth = $options['depth'];
         $this->preciseChildren = $options['precise_children'];
+
+        if (!$this->assetHelper instanceof CoreAssetsHelper) {
+            // check for the AssetsHelper class introduced in Symfony 2.7
+            if (!$this->assetHelper instanceof \Symfony\Bundle\FrameworkBundle\Templating\Helper\AssetsHelper) {
```
dbu: cool! please add a `use` statement for this class. this will not produce an error even if the class is not available. (sometimes the sloppiness of php is quite nice :-) )

wcluijt: The `use` statement has been added.

dbu:
thanks a lot!

taylankasap:
Why can't I get this commit when I do composer update with "sonata-project/doctrine-phpcr-admin-bundle": "~1.2" in composer.json. Is it because Travis CI failed?

dbu:
this is because we did not release a new 1.2 version since this was merged. i just tagged 1.2.3 now https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/releases/tag/1.2.3 so it should work.

taylankasap:
Thanks for the fast response. It works now.

PR #339, opened 2015-05-05
Cleanup for TreeBrowserBundle 2.0

WouterJ:
A cleanup of the bundle for the treebrowserbundle cleanup. Atm, the bundle is broken. This fixes it.

It contains some big BC breaks (removing tree for instance). Should I note this in the CHANGELOG file? (which doesn't contain any 2.0 changes atm)

Other things to be aware of:

* This now means that valid_children is no longer used. It's not very hard to implement it for the "create" action, but for displaying it requires new features in the resource bundle (/cc @dantleech am I correct or does it already have such feature?)
* Some options are now commented, I wanted to leave them in so we see what's still missing.
* Selected node is now no longer needed, as the treebrowser supports binding to the input field (automatically selecting based on the input field)


Controller/TreeController.php
```diff
@@ -22,9 +21,9 @@
 class TreeController extends Controller
```
dbu: do we still need such a controller in this bundle or will this all move to the tree bundle?

Tree/PhpcrOdmTree.php
```diff
@@ -1,470 +0,0 @@
...
- * A tree implementation to work with Doctrine PHPCR-ODM
- *
- * Your documents need to map all children with an Children mapping for the
- * tree to see its children. Not having the Children annotation is a
- * possibility to not show children you do not want to show.
- *
- * @author David Buchmann <david@liip.ch>
- * @author Uwe Jäger <uwej711@googlemail.com>
- */
-class PhpcrOdmTree implements TreeInterface
```
dbu: according to travis there is a test for this class that now produces a fatal error when running the tests...

dbu:
cool! i am happy to merge as soon as the tests are cleaned up. we should create issues for the currently abandoned functionality like valid_children however, so that we do not forget this.

and yes, lets add things to CHANGELOG.md and if possible also UPGRADE.md. doing this later will become much harder when we don't remember that well what we did.

Controller/TreeController.php
WouterJ: The controller is for the tree block, which is still used.

WouterJ:
Fixed the tests and updated the changelog.

dbu:
thanks a lot! as soon as it becomes clear, we should start writing a migration guide as well. but maybe wait until its more clear how the new things will work exactly.

WouterJ:
What do you mean with "a migration guide"?

dbu:
instructions how to upgrade a project from using admin 1.x to use admin 2.0. "upgrade guide" would be a better name probably.

PR #340, opened 2015-05-10
[WIP] Improve TreeBrowserBundle experience

wouterj:
This is probably one of the last big PRs regarding the TreeBrowserbundle 2.0. It changes the way the tree is used to something that's familair to all computer users.

When opening the form, the tree view is hidden and only the input box is shown:
cmf-tree-minimal

When the admin want to select a node, they can do 2 things:

* Type into the input field. The TreeBrowserBundle provides auto completion for the admin:
  cmf-tree-autocompletion
* Click the "select" button. This opens a tooltip with the tree:
   cmf-tree-tooltip

Both the tree view and the input field are keep in sync: Changing the input value changes the selected node in the tree and selecting a node in the tree changes the input value.

I'm not sure yet if it's very obvious what to do here for the admin, but I believe it is a step in the good direction as it'll hide those very long trees on a page.

This PR doesn't change admin for the manager (or admin) trees.

Fixes https://github.com/symfony-cmf/TreeBrowserBundle/issues/76

### DIFF
--- a/Resources/views/Form/form_admin_fields.html.twig
+++ b/Resources/views/Form/form_admin_fields.html.twig
@@ -158,17 +158,7 @@ file that was distributed with this source code.
                 {% set routing_default_values = routing_default_values|merge({default: app.request.attributes.get(default)}) %}
             {% endif %}
         {% endfor %}
-        var $treeTooltip = $('#{{ id }}-tree-tooltip');
-
-        $treeTooltip.hide(100);
-
-        $('#{{ id }}-tree-tooltip-open').click(function (e) {
-            e.preventDefault();
-
-            $treeTooltip.toggle();
-        });
-
-        $('#{{ id }}-tree-selector').cmfTree({
+        var tree = $('#{{ id }}-tree-selector').cmfTree({
             request: {
                 load: function (nodePath) {
                     return {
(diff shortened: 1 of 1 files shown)


Resources/views/Form/form_admin_fields.html.twig
```diff
@@ -172,15 +171,58 @@ file that was distributed with this source code.
...
 </script>
 
-<div id="{{id}}-tree-selector"></div>
+<div class="form-inline dropdown">
+    <input class="form-control" name="{{ full_name }}" id="{{ id }}-tree-selector-output" list="{{ id }}-tree-datalist" value="{{ value ?: (select_root_node ? root_node : '') }}">
+    <datalist id="{{ id }}-tree-datalist"></datalist>
 
-<div class="form-inline">
-    <input class="form-control" name="{{ full_name }}" id="{{ id }}-tree-selector-output" value="{{ value ?: (select_root_node ? root_node : '') }}">
+    <span class="btn btn-default btn-mini" data-toggle="dropdown">Select</span>
```
dbu: "Select" should be a translation. i suggest we could label this "Pick" instead wdyt?

Resources/views/Form/form_admin_fields.html.twig
```diff
@@ -172,15 +171,58 @@ file that was distributed with this source code.
...
+    <span class="btn btn-default btn-mini" id={{ id }}-tree-reset>{{ 'reset_tree' | trans({}, 'SonataDoctrinePHPCRAdmin') }}</span>
```
dbu: can we show a tooltip that says to what path the reset will reset?

dbu:
this is really cool! great work again @WouterJ , thanks a lot!

in what regard is this PR still WIP? or could we merge it already?

one thing i wonder is whether we should (in addition?) show the string representation of the selected tree node in addition to the path. we might also do that in the tree, show the nodename very small and the title bigger. but this would mean that the ResourceBundle would expose a technical id as well as the human id...  anyways, we should not block this PR by that idea.

WouterJ:
> in what regard is this PR still WIP? or could we merge it already?

I need to do some more testing, it's not bug free at the moment.

> one thing i wonder is whether we should (in addition?) show the string representation of the selected tree node in addition to the path. we might also do that in the tree, show the nodename very small and the title bigger.

As for the tree, it already shows the string representation (by default, using `PathUtil::getNodeName()` and using `Admin#toString()` when the sonata_admin enhancer is enabled).
For the input field, it only shows the path and not the string representation. I think we should not change this. (it's also the behaviour on Windows)

aitboudad:
:+1: ,  is there a way to only allow sub-tree of content only for example showing the whole tree when edition content doesn't make sense IMO.

WouterJ:
Yes, there is a `root_node` option for the `cmfTree` plugin that allows you to specify the root node.

hason:
:+1: Awesome!

dbu:
does the new browser also allow to say whether the root node should be selectable or not? lets say i want to select a route. i currently need to specify /cms so that /cms/routes (the root route) is selectable. i would want to specify /cms/routes and have an option to control whether the root is selectable or not.

Resources/views/Form/form_admin_fields.html.twig
```diff
@@ -172,15 +171,58 @@ file that was distributed with this source code.
...
+        $('#{{ id }}-tree-selector-output').cmfAutoComplete({
+            data: function (nodePath, next) {
+                if (tree.getFromCache(nodePath)) {
+                    next(tree.getFromCache(nodePath));
+
+                    return;
+                }
+
+                jQuery.ajax({
+                    url: '{{ path('_cmf_resource', {
```
hason: `_cmf_resource` -> `_cmf_get_resource`

OskarStark: ping @wouterj

wouterj:
Closing as the form type is moved to the TreeBrowserBundle: https://github.com/symfony-cmf/tree-browser-bundle/pull/116

PR #348, opened 2015-10-11
Fix Symfony deprecated feature usages

WouterJ:


Form/Type/ChoiceFieldMaskType.php
```diff
@@ -42,8 +43,14 @@ public function buildView(FormView $view, FormInterface $form, array $options)
      */
     public function setDefaultOptions(OptionsResolverInterface $resolver)
     {
-        parent::setDefaultOptions($resolver);
+        $this->configureOptions($resolver);
```
Soullivaneuh: Should have a comment with which version of Symfony is supported here.

lsmith77:
for 2.0 imho we should bump the minimum requirement to PHP 5.4

dbu:
can we just drop php 5.3 support now? it fails and imo its ok to give up a dead version of php in a minor version. if there are critical bugfixes, they can be backported to the last branch running on 5.3.

lsmith77:
like I said .. +1 from me

Soullivaneuh:
> can we just drop php 5.3 support now?

PHP version support dropping should be on new major.

lsmith77:
not sure I agree that it requires a new major but in this case we do have a new major version:
https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/blob/master/composer.json#L44

dbu:
php 5.4 does [not even have security support anymore](https://secure.php.net/supported-versions.php) - lets bump the version to 5.5 (like symfony 3 does).

dbu:
wouter, do you want to try dropping php 5.4 support?

patkar:
backref: sulu-io/sulu#2106

dbu:
@WouterJ would be great if you could rebase this and check the cs issue travis-ci is showing. i am ok to drop php 5.3 support if that keeps failing on travis.

Form/Type/ChoiceFieldMaskType.php
dbu: fixed in #363

dbu:
continued in #363

PR #349, opened 2015-11-04
Fix definition of Twig filters

hason:
```
An exception has been thrown during the compilation of a template
("Warning: Illegal offset type") in "layout.html.twig". 
```

```php
        foreach ($extension->getFilters() as $name => $filter) {
            if ($filter instanceof Twig_SimpleFilter) {
                $name = $filter->getName(); // $name is an array()!!
            }
            ...

            $this->filters[$name] = $filter;
        }
```


dbu:
ups. we seem to lack tests, this should not have happened. thanks for the fix!

PR #352, opened 2015-12-29
symfony-cmf/tree-browser-bundle v 2.0 doesn't exist

jmontoyaa:
According to 

https://github.com/symfony-cmf/TreeBrowserBundle


lsmith77:
But it will eventually:
https://github.com/symfony-cmf/TreeBrowserBundle/blob/master/composer.json#L26

Note that we only depend on this version in the 2.0 version of this Bundle:
https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/blob/master/composer.json#L44

PR #356, opened 2016-01-28
Use boolean instead of integer in BooleanFilter

billsworld:
Without this the query builder will output something like this ```WHERE (a.[published] = CAST('1' AS LONG)``` which does not work.


dbu:
thank you!

PR #357, opened 2016-01-29
require symfony/security-acl

greg0ire:


greg0ire:
@lsmith77 ready to merge :)

lsmith77:
thx

core23:
@greg0ire Sure that we need this dependency in the ``require`` part?

Other bundles added this to ``require-dev``: https://github.com/sonata-project/SonataAdminBundle/blob/master/composer.json

lsmith77:
@core23 good point .. can you submit a new PR?

greg0ire:
Not sure at all since I read [the error](https://travis-ci.org/sonata-project/SonataDoctrinePhpcrAdminBundle/jobs/105329673#L768) more closely : 

> PHP Fatal error:  Interface 'Symfony\Component\Security\Acl\Model\DomainObjectInterface' not found in /home/travis/build/sonata-project/SonataDoctrinePhpcrAdminBundle/vendor/sonata-project/admin-bundle/Admin/Admin.php on line 57

I think it should not be required at all, and sonata admin should require it…

greg0ire:
I just talked with @Soullivaneuh , and he told me that it is like this so that people may use the admin bundle without using its ACL functionality, and without requiring this package. He added we could also try to detect the presence of the package and act in a more sensible manner when it is not (return false in Admin::isAclEnabled() if acls are not configured, for instance.

See #358

greg0ire:
Guys, I just realized something :  since the `Admin` implements an ACL interface, anyone that tries to install sonata admin on sf >= 2.8 without also installing the ACL package will get an error even if he does not care about the ACL as soon as they create an Admin. Isn't that a bit crazy @core23 @OskarStark @rande @lsmith77 @Soullivaneuh ?

greg0ire:
If you agree please merge https://github.com/sonata-project/SonataAdminBundle/pull/3563 . After that, we won't need to require the ACL package in bundles that have nothing to do with it (not sure which one do not though)

greg0ire:
@core23 @lsmith77 : to be reverted when 2.3.8 is released, hopefully with this requirement

PR #358, opened 2016-01-30
require the ACL package only in dev

greg0ire:
It is not mentioned at all in the code, but the admin bundle, which is a
dependency of this package, mentions it in its code, which can make the
tests fail.


core23:
Thank you @greg0ire

PR #362, opened 2016-03-30
Travis: Cleanup configuration

core23:
- moved **php-cs-fixer** to travis
- **php-cs-fixers** runs once under PHP7
- **sphinx-build** runs once under PHP7
- added PHP nightly builds

Refs https://github.com/sonata-project/SonataAdminBundle/pull/3584

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,8 @@
 language: php
 
 php:
-  - 5.5
-  - 5.6
-  - 7.0
   - hhvm
+  - nightly
 
 sudo: false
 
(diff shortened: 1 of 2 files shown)


core23:
Some unit tests are failing. The last build is more than 2 months old

dbu:
i just restarted the master build and its still green, so it smells like something in this PR is breaking things.

.travis.yml
```diff
@@ -15,39 +13,50 @@ cache:
...
-    - php: 5.6
       env: SYMFONY_VERSION=2.7.*
     - php: 5.6
       env: SYMFONY_VERSION=2.8.*
+    - php: 5.5
+      env: SYMFONY_VERSION=3.0.*
     - php: 5.6
       env: SYMFONY_VERSION=3.0.*
+    - php: 7.0
+      env: SYMFONY_VERSION=3.0.*
```
dbu: this is a lot of builds. i would say we want to build every php version once and every symfony version once with a specific php version. no need to build symfony 3 3 times.

dbu:
tough i can't see why that would be. i guess rebasing this branch does not change anything?

.travis.yml
core23: That was a long discussion in the past weeks what we should test in travis... Too many environments shouldn't be any problem ;)

Only symfony 3.0 is build by all supported php versions. After the next major release of ``Sonata*Bundle`` the build will be very short.

core23:
Rebased this and everything is green @dbu

dbu:
\o/

PR #363, opened 2016-04-01
Fix Symfony deprecated feature usages

dbu:
continuation of #348

PR #364, opened 2016-04-01
fix code quality issues in tests

dbu:
fix things noticed in #363

### DIFF
--- a/Tests/Builder/ListBuilderTest.php
+++ b/Tests/Builder/ListBuilderTest.php
@@ -11,10 +11,45 @@
 
 namespace Sonata\DoctrinePHPCRAdminBundle\Tests\Builder;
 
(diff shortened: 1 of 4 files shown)


Tests/Builder/ListBuilderTest.php
```diff
@@ -42,9 +77,6 @@ private function setupAddField()
 
         $this->fieldDescription = $this->getMock('\Sonata\AdminBundle\Admin\FieldDescriptionInterface');
         $this->fieldDescription->expects($this->any())
-            ->method('getModelManager')
```
dbu: this method does not exist. if its called, we would get a phpunit mock error.

dbu:
yay, at least SonataCI is happy now :-)

PR #367, opened 2016-04-09
Fix even more Symfony 3 support (and some other bugs)

WouterJ:
Tests still have to be updated. This was the minimum required things to get at least something working. My guess is that this bundle needs a lot more fixes before working 100% on Symfony 3.

Can we bump to PHP 5.5 support, or do I still need to put the FQCN in strings instead of using `::class`?

### DIFF
--- a/Guesser/FilterTypeGuesser.php
+++ b/Guesser/FilterTypeGuesser.php
@@ -11,15 +11,15 @@
 
 namespace Sonata\DoctrinePHPCRAdminBundle\Guesser;
 
+use Doctrine\Bundle\PHPCRBundle\Form\Type\DocumentType;
 use Doctrine\Bundle\PHPCRBundle\ManagerRegistry;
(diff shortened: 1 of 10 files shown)


Guesser/FilterTypeGuesser.php
```diff
@@ -62,7 +66,7 @@ public function guessType($class, $property, ModelManagerInterface $modelManager
             $options['operator_type'] = 'sonata_type_boolean';
             $options['operator_options'] = array();
 
-            $options['field_type'] = 'document';
+            $options['field_type'] = DocumentType::class;
```
WouterJ: `document` never existed afaik. I guess `phpcr_document` was meant to be used (so in fact, nobody ever used this filter)

dbu:
cool, great job wouter.

yeah, lets drop php < 5.5

according to travis, you missed one case of the TextType.

WouterJ:
Fixed tests and CS. Turns out PHP <5.5 was already dropped.

PR #370, opened 2016-05-02
Composer improvements

Soullivaneuh:
* 2.x-dev
* Use new composer constraints notation
* Remove minimum stability

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,33 +16,30 @@
         }
     ],
     "require": {
-        "php": "~5.5|~7.0",
-        "sonata-project/admin-bundle": "~3.0",
-        "sonata-project/block-bundle": "^2.3.2",
-        "symfony/framework-bundle": "~2.3|~3.0",
-        "symfony/property-access": "~2.2|~3.0",
-        "doctrine/phpcr-odm": "~1.1",
-        "doctrine/phpcr-bundle": "~1.1",
-        "symfony-cmf/tree-browser-bundle": "~2.0",
-        "symfony-cmf/resource-rest-bundle": "~1.0"
+        "php": "^5.5 || ^7.0",
+        "sonata-project/admin-bundle": "3.x-dev@dev",
+        "symfony/framework-bundle": "^2.3 || ^3.0",
+        "symfony/property-access": "^2.2 || ^3.0",
+        "doctrine/phpcr-odm": "^1.1",
+        "doctrine/phpcr-bundle": "^1.1",
+        "symfony-cmf/tree-browser-bundle": "^2.0@dev",
+        "symfony-cmf/resource-rest-bundle": "^1.0@dev"
     },
     "require-dev": {
-        "jackalope/jackalope-jackrabbit": "~1.0",
-        "symfony/phpunit-bridge": "~2.7|~3.0",
-        "symfony/security-acl": "~2.4|~3.0",
-        "fabpot/php-cs-fixer": "~0.1|~1.0"
+        "jackalope/jackalope-jackrabbit": "^1.0",
+        "symfony/phpunit-bridge": "^2.7 || ^3.0",
+        "symfony/security-acl": "^2.4 || ^3.0",
+        "fabpot/php-cs-fixer": "~0.1 || ^1.0"
     },
     "provide": {
         "sonata-project/admin-bundle-persistency-layer": "1.0.0"
     },
-    "minimum-stability": "dev",
-    "prefer-stable": true,
     "autoload": {
         "psr-4": { "Sonata\\DoctrinePHPCRAdminBundle\\": "" }
     },
     "extra": {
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -16,33 +16,30 @@
...
-        "sonata-project/admin-bundle": "~3.0",
-        "sonata-project/block-bundle": "^2.3.2",
-        "symfony/framework-bundle": "~2.3|~3.0",
-        "symfony/property-access": "~2.2|~3.0",
-        "doctrine/phpcr-odm": "~1.1",
-        "doctrine/phpcr-bundle": "~1.1",
-        "symfony-cmf/tree-browser-bundle": "~2.0",
-        "symfony-cmf/resource-rest-bundle": "~1.0"
+        "php": "^5.5 || ^7.0",
+        "sonata-project/admin-bundle": "3.x-dev@dev",
```
dbu: same question here, why not 3.*?

composer.json
```diff
@@ -16,33 +16,30 @@
...
+        "symfony/framework-bundle": "^2.3 || ^3.0",
+        "symfony/property-access": "^2.2 || ^3.0",
+        "doctrine/phpcr-odm": "^1.1",
+        "doctrine/phpcr-bundle": "^1.1",
+        "symfony-cmf/tree-browser-bundle": "^2.0@dev",
```
dbu: this is actually different than having the minimum-stability: we make every user install dev versions of the tree bundle even once a stable version of the tree bundle is out. assuming that we don't release before tree bundle 2.0 is stable, we can still correct that however.

greg0ire: Should we use `^2.0@dev|^2.0` then ?

dbu: that makes no difference, it still opens the dependency to dev versions. 

in the cmf, we preferred to use the minimum-stability during development and remove that before releaseing (when the dependencies hopefully have stable versions around). actually keeping minimum-stability can reveil accidental BC breaks in the dependencies. as it only affects composer run in this repository and not the including project, i find it cleaner than adding `@dev` into dependencies. but if somebody already starts using this, they need to include all our deps explicitly or also set minimum-stability in the project to end up with something installable. so really not sure which is the more elegant approach.

greg0ire: IMO adding `@dev` to some dependencies is cleaner because explicit, and seems to make Composer faster, probably because it reduces possibilities.

> that makes no difference, it still opens the dependency to dev versions. 

Maybe it is best to wait then, that way we don't forget to remove this `@dev` when a stable release is out.

WouterJ: I agree that `@dev` is cleaner than the minimum-stability. Please note that the same rules apply to the tags as to the min stability (both are root-only).

In the CMF, I switched to min stability because there were a lot of dependencies that we don't really depend on that needed a dev stability (because of deps of deps).

composer.json
WouterJ: This should at least be `^3.0@dev` and not `3.x-dev@dev` (which is double btw, `3.x-dev` should be enough).

composer.json
```diff
@@ -16,33 +16,30 @@
...
-        "php": "~5.5|~7.0",
-        "sonata-project/admin-bundle": "~3.0",
-        "sonata-project/block-bundle": "^2.3.2",
-        "symfony/framework-bundle": "~2.3|~3.0",
-        "symfony/property-access": "~2.2|~3.0",
-        "doctrine/phpcr-odm": "~1.1",
-        "doctrine/phpcr-bundle": "~1.1",
-        "symfony-cmf/tree-browser-bundle": "~2.0",
-        "symfony-cmf/resource-rest-bundle": "~1.0"
+        "php": "^5.5 || ^7.0",
```
WouterJ: any reason to use `||` instead of the quite common `|`?

WouterJ:
LGTM, CMF is blocked by this atm.

composer.json
Soullivaneuh: > instead of the quite common `|`?

This is the problem: it shouldn't.

`||` is the new standard. See examples of docs: https://getcomposer.org/doc/04-schema.md

Soullivaneuh:
@WouterJ Just two points:

* The working branch of this PR is an unstable one. I can understand your project is stuck but an unstable branch should be used for production or stable dependent releases.
* The goal of "improving composer" is part of beginning new sonata version management with a good start. It's maybe more painful but necessary IMHO.

composer.json
greg0ire: `3.x-dev@dev` looks indeed fishy…

composer.json
WouterJ: Thanks for the info, good to know that the single one is actually sort-of deprecated.

WouterJ:
see https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/pull/371

PR #371, opened 2016-05-04
Fix composer dependencies

WouterJ:
*I know I'm not allowed to use the word "fix", but I can't come up with a better word quickly here, sorry.*

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,8 +34,7 @@
         "sonata-project/core-bundle": "@dev",
         "jackalope/jackalope-jackrabbit": "^1.0",
         "symfony/phpunit-bridge": "^2.7 || ^3.0",
-        "symfony/security-acl": "^2.4 || ^3.0",
-        "fabpot/php-cs-fixer": "^1.0"
+        "symfony/security-acl": "^2.4 || ^3.0"
     },
     "provide": {
         "sonata-project/admin-bundle-persistency-layer": "1.0.0"
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,8 @@
     ],
     "require": {
         "php": "^5.5 || ^7.0",
-        "sonata-project/admin-bundle": "3.x-dev@dev",
+        "sonata-project/admin-bundle": "^3.0@dev",
+        "sonata-project/block-bundle": "^3.0@dev",
         "symfony/framework-bundle": "^2.3 || ^3.0",
         "symfony/property-access": "^2.2 || ^3.0",
         "doctrine/phpcr-odm": "^1.1",
@@ -26,10 +27,15 @@
         "symfony-cmf/resource-rest-bundle": "^1.0@dev"
     },
     "require-dev": {
+        "symfony-cmf/resource-bundle": "@dev",
+        "symfony-cmf/resource": "@dev",
+        "dantleech/glob-finder": "@dev",
+        "puli/repository": "@beta",
+        "sonata-project/core-bundle": "@dev",
         "jackalope/jackalope-jackrabbit": "^1.0",
         "symfony/phpunit-bridge": "^2.7 || ^3.0",
         "symfony/security-acl": "^2.4 || ^3.0",
-        "fabpot/php-cs-fixer": "~0.1 || ^1.0"
+        "fabpot/php-cs-fixer": "^1.0"
     },
     "provide": {
         "sonata-project/admin-bundle-persistency-layer": "1.0.0"
--- a/composer.json
+++ b/composer.json
@@ -16,33 +16,30 @@
         }
     ],
     "require": {
-        "php": "~5.5|~7.0",
-        "sonata-project/admin-bundle": "~3.0",
-        "sonata-project/block-bundle": "^2.3.2",
-        "symfony/framework-bundle": "~2.3|~3.0",
-        "symfony/property-access": "~2.2|~3.0",
-        "doctrine/phpcr-odm": "~1.1",
-        "doctrine/phpcr-bundle": "~1.1",
-        "symfony-cmf/tree-browser-bundle": "~2.0",
-        "symfony-cmf/resource-rest-bundle": "~1.0"
+        "php": "^5.5 || ^7.0",
+        "sonata-project/admin-bundle": "3.x-dev@dev",
+        "symfony/framework-bundle": "^2.3 || ^3.0",
+        "symfony/property-access": "^2.2 || ^3.0",
+        "doctrine/phpcr-odm": "^1.1",
+        "doctrine/phpcr-bundle": "^1.1",
+        "symfony-cmf/tree-browser-bundle": "^2.0@dev",
+        "symfony-cmf/resource-rest-bundle": "^1.0@dev"
     },
     "require-dev": {
-        "jackalope/jackalope-jackrabbit": "~1.0",
-        "symfony/phpunit-bridge": "~2.7|~3.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -16,33 +16,36 @@
...
     "require-dev": {
-        "jackalope/jackalope-jackrabbit": "~1.0",
-        "symfony/phpunit-bridge": "~2.7|~3.0",
-        "symfony/security-acl": "~2.4|~3.0",
-        "fabpot/php-cs-fixer": "~0.1|~1.0"
+        "symfony-cmf/resource-bundle": "@dev",
+        "symfony-cmf/resource": "@dev",
+        "dantleech/glob-finder": "@dev",
+        "puli/repository": "@beta",
+        "sonata-project/core-bundle": "@dev",
```
WouterJ: see here the problem of not using `minimum-stability` in this case: A lot of packages that other packages depend on need to be whitelisted for dev.

greg0ire: Indeed… and there is a lot of them :confused:

composer.json
```diff
@@ -16,33 +16,36 @@
...
+        "jackalope/jackalope-jackrabbit": "^1.0",
+        "symfony/phpunit-bridge": "^2.7 || ^3.0",
+        "symfony/security-acl": "^2.4 || ^3.0",
+        "fabpot/php-cs-fixer": "^1.0"
```
WouterJ: Remove `~0.1`, as it makes no sense to install an older version for such a tool (it'll only make Composer slower in resolving versions).

*Btw, I don't see any reason to add this as a project dependency as nothing relies on this*

composer.json
```diff
@@ -16,33 +16,36 @@
...
     },
-    "minimum-stability": "dev",
-    "prefer-stable": true,
     "autoload": {
         "psr-4": { "Sonata\\DoctrinePHPCRAdminBundle\\": "" }
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0-dev"
+            "dev-master": "2.x-dev"
```
WouterJ: I don't understand this change, but the change was part of #370 so I left it here.

composer.json
Soullivaneuh: > Remove `~0.1`, as it makes no sense to install an older version for such a tool (it'll only make Composer slower in resolving versions).

This will not works on lowest dependencies, that's why.

> *Btw, I don't see any reason to add this as a project dependency as nothing relies on this*

This is the old way on it gonna to be changes like we did in https://github.com/sonata-project/SonataAdminBundle/pull/3273.

Soullivaneuh:
> I know I'm not allowed to use the word "fix", but I can't come up with a better word quickly here, sorry.

That's @greg0ire said. :stuck_out_tongue:

composer.json
Soullivaneuh: https://github.com/sonata-project/SonataAdminBundle/issues/3053

composer.json
Soullivaneuh: Maybe waiting new release management, but this have to be fixed later.

greg0ire:
> I know I'm not allowed to use the word "fix", but I can't come up with a better word quickly here, sorry.

You can totally use it, but in the description of your commit message (which is why you are doing this). The commit message subject should describe what you are doing. From what I see in the diff you are : 

- whitelisting some indirect dependencies
- adding a dependency on the block bundle
- changing the name of the version constraint of the admin bundle

This means there should be 3 "well-scoped" commits. I totally agree with you on that part, a commit named "Composer improvements" is not great, but I'm trying to give @Soullivaneuh a break because I know he has to do that on every bundle, which could be painful.

composer.json
WouterJ: Discovered it was already globally installed by the Travis config. So a project dep was not needed, removed the dep completely.

.travis.yml
```diff
@@ -43,20 +43,22 @@ matrix:
...
-  - travis_wait composer update --prefer-dist --no-interaction $COMPOSER_FLAGS
   - export PATH=$HOME/.local/bin:$PATH
-  - pip install -r Resources/doc/requirements.txt --user `whoami`
 
-script:
- - make $TARGET
+install:
+  - if [ "$TARGET" != "docs" ]; composer global update --prefer-dist --no-interaction; fi;
+  - if [ "$TARGET" = "test" ]; travis_wait composer update --prefer-dist --no-interaction $COMPOSER_FLAGS; fi;
+  - if [ "$TARGET" = "docs" ]; pip install -r Resources/doc/requirements.txt --user `whoami`; fi;
```
greg0ire: Could be grouped with the first statement, like this : 

```yml
- | 
    if [ "$TARGET" = "docs" ]; then
    pip install -r Resources/doc/requirements.txt --user `whoami`
    else
    composer global update --prefer-dist --no-interaction; fi;
    fi
```

WouterJ: I prefer it this way, as the 2 actions are not really related. It's more like: If it's "test" or "cs-fix", run a global update. If it's docs run a pip install.

As `if [[ "$TARGET" = "test" || "$TARGET" = "cs_dry_run" ]]; then ...; fi;` is more complex than `if [ "$TARGET" != "docs" ]; then ...; fi;`.

WouterJ:
Build passes :balloon:

.travis.yml
greg0ire: As you wish.

.travis.yml
```diff
@@ -43,20 +43,22 @@ matrix:
...
+  - if [ "$TARGET" = "test" ]; then composer global require phpunit/phpunit:@stable --no-update; fi;
   - if [ "$SYMFONY_VERSION" != "" ]; then composer require "symfony/symfony:${SYMFONY_VERSION}" --no-update; fi;
-  - travis_wait composer update --prefer-dist --no-interaction $COMPOSER_FLAGS
   - export PATH=$HOME/.local/bin:$PATH
-  - pip install -r Resources/doc/requirements.txt --user `whoami`
 
-script:
- - make $TARGET
+install:
+  - if [ "$TARGET" != "docs" ]; then composer global update --prefer-dist --no-interaction; fi;
```
Soullivaneuh: Lot of repetitions for no real benefits IMO.

Requiring two packages globally even if it's not used is not painful, could you please revert this part?

BTW, `.travis.yml` would be a generated file by dev-kit at the end: https://github.com/sonata-project/dev-kit/pull/31

Soullivaneuh: And BTW, like as said, the `cs_dry_run` will not be here later.

WouterJ: Reverted all Travis CI changes

Soullivaneuh:
@WouterJ Could you please elaborate your "Fix build" commit message?

.travis.yml
WouterJ: Please note that it made quite a difference:

 * **17min 11sec** - Test with these changes (https://travis-ci.org/sonata-project/SonataDoctrinePhpcrAdminBundle/builds/127749164)
 * **22min 24sec** - Test without these changes (https://travis-ci.org/sonata-project/SonataDoctrinePhpcrAdminBundle/builds/127751689)

greg0ire:
> @WouterJ Could you please elaborate your "Fix build" commit message?

Don't bother, I'll do it for you since you seem quite busy with other things.

WouterJ:
Pushed a new commit with a better commit message.

greg0ire:
Thanks a lot @WouterJ @Soullivaneuh

PR #375, opened 2016-06-01
Use render() function instead of block

WouterJ:
Another step towards Symfony 3 compatibility.

### DIFF
--- a/Resources/views/Block/tree.html.twig
+++ b/Resources/views/Block/tree.html.twig
@@ -17,7 +17,10 @@ file that was distributed with this source code.
             <h3 class="box-title">{{ 'content_tree' | trans({}, 'SonataDoctrinePHPCRAdmin') }}</h3>
         </div>
         <div class="box-body">
-            {{ render(controller('sonata.admin.doctrine_phpcr.tree_controller:treeAction', { 'root': settings.id, 'selected': settings.selected })) }}
+            {{ render(controller('sonata.admin.doctrine_phpcr.tree_controller:treeAction', {
+                'root': settings.id,
+                'selected': settings.selected
+            })) }}
         </div>
     </div>
 {% endblock %}
--- a/Resources/views/Block/tree.html.twig
+++ b/Resources/views/Block/tree.html.twig
@@ -17,7 +17,7 @@ file that was distributed with this source code.
             <h3 class="box-title">{{ 'content_tree' | trans({}, 'SonataDoctrinePHPCRAdmin') }}</h3>
         </div>
         <div class="box-body">
-            {% render controller('sonata.admin.doctrine_phpcr.tree_controller:treeAction', { 'root': settings.id, 'selected': settings.selected }) %}
+            {{ render(controller('sonata.admin.doctrine_phpcr.tree_controller:treeAction', { 'root': settings.id, 'selected': settings.selected })) }}
         </div>
     </div>
 {% endblock %}


Soullivaneuh:
@WouterJ Is that compatible with SF 2.3?

WouterJ:
@Soullivaneuh yes: https://github.com/symfony/twig-bridge/blob/2.3/Extension/HttpKernelExtension.php#L39

Resources/views/Block/tree.html.twig
```diff
@@ -17,7 +17,7 @@ file that was distributed with this source code.
             <h3 class="box-title">{{ 'content_tree' | trans({}, 'SonataDoctrinePHPCRAdmin') }}</h3>
         </div>
         <div class="box-body">
-            {% render controller('sonata.admin.doctrine_phpcr.tree_controller:treeAction', { 'root': settings.id, 'selected': settings.selected }) %}
+            {{ render(controller('sonata.admin.doctrine_phpcr.tree_controller:treeAction', { 'root': settings.id, 'selected': settings.selected })) }}
```
greg0ire: Would you consider also linebreaking this in a separate commit while you're at it ?

WouterJ: Done

greg0ire: Thanks a lot for doing that, another step towards sane code.

WouterJ:
Tests fail because this bundle's requirements no longer support Symfony 2.3. What about dropping it? (Symfony 2.3 has reached end of maintainance some days ago)

greg0ire:
Or maybe find a way to make it BC? @Soullivaneuh to the rescue ? BTW, if this was planned to be BC, should it have been based on master? Why is master the default branch for this bundle ? It not very clear what the status is even after reading https://github.com/sonata-project/SonataAdminBundle/issues/3731

WouterJ:
@greg0ire it's about the Resource(Rest)Bundle dependency. This bundle's 1.0 version is going to be included in the CMF 2 release, which targets Symfony 2.8+ and PHP 5.5+.

As this bundle depends on it, this bundle's 2.0 version (master is 2.0 dev) will also only support 2.8+ and PHP 5.5+.

greg0ire:
I think it is safe to drop 2.3 . ping @dbu @lsmith77

WouterJ:
Btw, this is not an issue caused by this PR, master is broken by default. Can we please discuss this in another pull request and merge this one?

greg0ire:
The build is green for master : https://travis-ci.org/sonata-project/SonataDoctrinePhpcrAdminBundle/builds/129346090 , so I'm very reluctant to merge this. It would mean we would have to check every job of every build to make sure the only tests that are failing are the one that fail in this PR. Besides, I think dropping 2.3 support can be done quite quickly.

greg0ire:
Oh sorry I misread your post, master should indeed be broken.

greg0ire:
Well since there is no current way to make it green, I guess I'll just merge.

greg0ire:
@Soullivaneuh , in [this file](https://github.com/Soullivaneuh/sonata-dev-kit/blob/master/config/projects.yml#L142), does the `~` mean phpcr does not use a dev-kit-generated composer.json yet ? If yes are we supposed to drop support by doing a direct PR, or should be take that as an occasion to fill in the blanks in dev-kit ?

WouterJ:
Thanks for the merge!

greg0ire:
> Thanks for the merge!

Thanks for the contribution ;)

Soullivaneuh:
> does the ~ mean phpcr does not use a dev-kit-generated composer.json

First of all, dev-kit does not generate composer.json. It distpatch a lint.

This bundle is one of the projects I still have to integrate.

And yes, all projects with no configuration (`~`) are project not integrated with dev-kit **for the files distpatcher**.

PR #377, opened 2016-06-03
[WIP] Update bundle for move ResourceRestBundle changes

wouterj:
Implements drag 'n drop support in this bundle. It also fixes the renamed route names. 

To be merged once https://github.com/symfony-cmf/resource-rest-bundle/pull/31 is merged

Depends on
---

 * [ ] https://github.com/symfony-cmf/tree-browser-bundle/pull/99
 * [x] https://github.com/symfony-cmf/resource-rest-bundle/pull/36

### DIFF
--- a/Resources/views/Tree/tree.html.twig
+++ b/Resources/views/Tree/tree.html.twig
@@ -24,6 +24,7 @@ file that was distributed with this source code.
 <script type="text/javascript">
     jQuery(function($) {
         $('#tree').cmfTree({
+            dnd: { enabled: true },
             request: {
                 load: function (nodePath) {
                     return {
@@ -33,6 +34,13 @@ file that was distributed with this source code.
                             path: '__path__'
                         }|merge(routing_default_values)) }}'.replace('__path__', nodePath)
                     };
+                },
+                move: function (nodePath) {
+                    return '{{ path('_patch_cmf_resource', {
+                        repositoryName: repository_name,
+                        root_node: root_node,
+                        path: '__path__'
+                    }|merge(routing_default_values)) }}'.replace('__path__', nodePath)
                 }
             },
             actions: {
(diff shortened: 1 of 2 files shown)


dbu:
looks good to me. seems to no longer install with symfony 2.3 (i guess this is not related to the PR, so we can still merge).

not sure if sonataCI is red because of the 2.3 install - the dashboard only tells me everything is fine, don't see why its red.

Resources/views/Tree/tree.html.twig
```diff
@@ -24,15 +24,23 @@ file that was distributed with this source code.
...
     jQuery(function($) {
         $('#tree').cmfTree({
+            dnd: { enabled: true },
             request: {
                 load: function (nodePath) {
                     return {
-                        url: '{{ path('_cmf_resource', {
+                        url: '{{ path('_get_cmf_resource', {
                             repositoryName: repository_name,
                             root_node: root_node,
```
dbu: is the github parser not capable of handling multiline strings? oh, or is this actually wrong and would need quotes? i think this assumes its twig variables, no?

WouterJ: It's confused because we're mixing javascript and twig. It doesn't understand twig and simply assumes it's HTML (with javascript included).

These big red warnings can simply be ignored.

dbu:
@WouterJ ah right, json in javascript in twig... what is still WIP about this PR? shall we merge it? i think the errors are not related to the PR, as it builds fine on all but symfony 2.3 and the diff does not affect symfony versions.

WouterJ:
@dbu updated the PR descriptions with PRs that have to be merged first.

Clearly, this isn't covered by any tests at the moment, as both the tree form type as the tree blocks should result in an exception (the route does not exists yet in ResourceRestbundle master).

core23:
Any updates here? @wouterj

wouterj:
the form type is moved to the CMF treebrowserbundle. Will update this bundle soon to deprecate/remove most of the tree stuff.

PR #381, opened 2016-06-04
DevKit updates for master branch

SonataCI:


Soullivaneuh:
@dbu @WouterJ master branch seems to be not compatible with Symfony 2.3. Is that normal?

WouterJ:
@Soullivaneuh Symfony CMF 2.0 is targetting Symfony 2.8+ (as we don't have much resources as well, we decided it's impossible to maintain tons of BC layers).

This bundle requires the TreeBrowserBundle 2.0 and Resource* packages (both part of CMF 2.0), so the dependencies don't work on Symfony 2.3 and thus this bundle doesn't as well.

However, I have plans to move the tree form type to the CmfTreeBrowserbundle (https://github.com/symfony-cmf/tree-browser-bundle/issues/87). If that happends, this bundle no longer need to depend on TreeBrowserBundle and Resource* bundles and can regain Symfony 2.3 support.

Soullivaneuh:
> However, I have plans to move the tree form type to the CmfTreeBrowserbundle (symfony-cmf/tree-browser-bundle#87). If that happends, this bundle no longer need to depend on TreeBrowserBundle and Resource* bundles and can regain Symfony 2.3 support.

Ok so what should I do for now? Drop 2.3 support? Because Travis is crying right now. :-)

WouterJ:
> Ok so what should I do for now? Drop 2.3 support? Because Travis is crying right now. :-)
 Hide all checks

I'm not sure about what your support plan is for Sonata. If all other new sonata majors will support Symfony 2.3, it would make sense to support it here as well. You can add Symfony 2.3 jobs to allowed_failures for the few weeks it'll be incompatible (or comment them out).

Soullivaneuh:
> If all other new sonata majors will support Symfony 2.3, it would make sense to support it here as well.

Actual major does support Symfony 2.3. We are planning to drop it on next one.

See https://github.com/sonata-project/dev-kit/issues/51 and https://github.com/sonata-project/dev-kit/issues/52.

> You can add Symfony 2.3 jobs to allowed_failures for the few weeks it'll be incompatible (or comment them out).

Dev-kit is not ready for that right now. :-/

Will take a look.

After that, if you want to drop Symfony 2.3 right now for this bundle, I don't see any inconvenient.

dbu:
i'd say lets just drop 2.3, its EOL and if people upgrade the sonata suite they hopefully also upgrade symfony.

Soullivaneuh:
Ok so let specify it on the composer.json of this bundle.

composer.json
```diff
@@ -23,8 +23,8 @@
         "sonata-project/block-bundle": "^3.0",
         "symfony-cmf/resource-rest-bundle": "^1.0@dev",
         "symfony-cmf/tree-browser-bundle": "^2.0@dev",
-        "symfony/framework-bundle": "^2.3 || ^3.0",
-        "symfony/property-access": "^2.2 || ^3.0"
+        "symfony/framework-bundle": "^2.8 || ^3.0",
+        "symfony/property-access": "^2.8 || ^3.0"
```
dbu: why not 2.7 for the LTS?

Soullivaneuh: Two reason:

* As @WouterJ said, Symfony CMF 2.0 is targetting Symfony 2.8+.
* The next major for other Sonata project will target SF 2.8 that is a LTS too: https://github.com/sonata-project/dev-kit/issues/51. So if we drop Symfony 2.3 right now, let's start on 2.8.

Issue #373, opened 2016-05-10
tree-browser and resource-rest stable ETA


soullivaneuh:
Have you an ETA for `symfony-cmf/tree-browser-bundle` and `symfony-cmf/resource-rest-bundle` stable release?

I would like to push a new major for sonata-admin 3.0 support and deprecate the 1.x one but those bundle are already required with unstables versions.

Thanks.

cc dbu @lsmith77 @dantleech @sjopet @rmsint

dbu:
@WouterJ is most active with the tree browser. can you please check the milestones and make sure the relevant issues that need to be done are there? and maybe label things as nicetohave that we could also do without?

WouterJ:
The biggest "problem" is Puli. It has to be released stable in order to mark ResourceRestbundle stable.

WouterJ:
ping @webmozart, what are the plans to release Puli 1.0 stable?

iBasit:
+1 

it will really help to get basic stable versions for all the chained bundles

Soullivaneuh:
I will start dev-kit integration.

1.x will be the stable branch for now, but not compatible with the new ecosystem.

When all dependencies will be stable on master, we will release a 2.x tag and deprecate the 1.x.

dbu:
thanks!

wouterj:
Fyi, we're now about to release RC's of CMF 2.0. Expect a stable release this month.

PR #392, opened 2016-06-21
Support Doctrine PHPCR-ODM 2.0

WouterJ:
Doctrine PHPCR-ODM 2.0-dev has started, which will be required for the Symfony CMF 2.0-dev.

At the moment, no breaking changes affect the features in this bundle, so both 1.x and 2.x can be supported now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "php": "^5.5 || ^7.0",
         "doctrine/phpcr-bundle": "^1.1",
-        "doctrine/phpcr-odm": "^1.1",
+        "doctrine/phpcr-odm": "^1.1 || ^2.0",
         "sonata-project/admin-bundle": "^3.1",
         "sonata-project/block-bundle": "^3.0",
         "symfony-cmf/resource-rest-bundle": "^1.0@dev",


Soullivaneuh:
It would be great to wait a stable version before merge it.

No stable version of this bundle is currently out because of unstable dependencies.

Ref: #373

cc @sonata-project/contributors

WouterJ:
> It would be great to wait a stable version before merge it.

This would mean all CMF development stops. It would require PHPCR ODM 2.0, but as the Sonata admin doesn't support this version, work cannot continue. PHPCR ODM 2.0 mainly relies on CMF development, so it kinda stops development as well and this bundle relies on CMF 2.0 and PHPCR ODM 2.0 development (you guess it, that stops as well).

I like how you want to keep Sonata as stable as it can get, but we need to have some room to work on features. In this complex world of packages relying on many third party packages (one of the problems of both CMF and Sonata), this means that dev branches really are unstable.

composer.json
```diff
@@ -18,7 +18,7 @@
     "require": {
         "php": "^5.5 || ^7.0",
         "doctrine/phpcr-bundle": "^1.1",
-        "doctrine/phpcr-odm": "^1.1",
+        "doctrine/phpcr-odm": "^1.1 || ^2.0",
```
Soullivaneuh: Should not be `^2.0@dev`?

Soullivaneuh:
Well if this is really necessary then go for it.

In this case we should wait stabilization of dependencies, then release a 2.0 and next, take care of not using unstable dependencies for the future. :wink:

composer.json
WouterJ: If I understand Composer correctly, the `@dev` flag applies to the complete constraint: `(^1.1 || ^2.0)@dev`.

Also, the dev flag does apply to root-only. So packages needing PHPCR-ODM 2.0 would need to set the minimum stability to dev anyway.

If we want to test against this, I suggest adding a dev-dependency test job like this: https://github.com/symfony-cmf/routing-bundle/blob/master/.travis.yml#L35

WouterJ:
> In this case we should wait stabilization of dependencies, then release a 2.0 and next, take care of not using unstable dependencies for the future. 😉

Yeah, probably. The good news is that Sonata is going to be extracted from the CMF core bundles, so I'll stop pushing Sonata's policy in a few months 😉

Soullivaneuh:
> The good news is that Sonata is going to be extracted from the CMF core bundles

What do you mean? Not using it for your project? :cry:

WouterJ:
> What do you mean? Not using it for your project? 😢

At the moment, the Symfony CMF main bundles provide admins in their bundle: https://github.com/symfony-cmf/routing-bundle/tree/master/Admin This means we can't release a new version of a main bundle without having a working admin.

With CMF 2.0, we extract these admin classes to a SonataAdminIntegration bundle: https://github.com/symfony-cmf/sonata-admin-integration-bundle This means we can release new versions of main bundles without having a working admin. (as the admin integration can be released independently). This means we don't have to push Sonata admin development as much as we do now in order to keep CMF dev going.

Soullivaneuh:
Well, you will have the same "problem" with SonataAdminIntegration bundle in this case. Am I wrong?

WouterJ:
> Well, you will have the same "problem" with SonataAdminIntegration bundle in this case. Am I wrong?

Yeah, except that we make things a lot less complex. In the current situation:

 * SonataPhpcrAdmin requires working CMF packages
 * CMF packages requires working SonataPhpcrAdmin package

This means both have to work with unstable dependencies.

In the new situation:

 * SonataPhpcrAdmin requires working main CMF packages
 * Main CMF packages no longer requires sonata
 * A special sonata integration package requires working main CMF packages and main Sonata packages

This means both Sonata and CMF packages can work with stable dependencies at all times. (we would for instance first release stable CMF main packages, then a stable sonata package and then work on a stable sonata integration)

Soullivaneuh:
Oh yes, you are avoiding a circular dependency right?

Didn't see this package had something like this...

We have also two other circular dependencies under sonata packages. Would be great to get rid of this sh*t too... :stuck_out_tongue:

composer.json
dbu: afaik `@dev` is used by others depending on this bundle. not using `@dev` is the right thing here. projects that happen to use master of sonata admin will still get the stable phpcr-odm unless they either specifically ask for the 2 version `@dev` or change minimum-stability.

dbu:
@WouterJ can you try to re-push a commit to trigger travis-ci? it looks like there was a glitch and it missed the build.

WouterJ:
Pushed again

dbu:
@Soullivaneuh is the sonata ci thing red because of the phpunit warnings about getMock vs createMock? if so thats not related to the PR and we can merge.

Soullivaneuh:
> @Soullivaneuh is the sonata ci thing red because of the phpunit warnings about getMock vs createMock? if so thats not related to the PR and we can merge.

No, this is just warnings.

Soullivaneuh:
Oh you are talking SonataCI... You can ignore it, it should be removed BTW (cc @rande)

dbu:
thanks wouter!

PR #408, opened 2016-11-04
Seems to add the correct checkboxes to the form

syeo66:
I am targetting this branch, because it is just a small addition to the Ajax callback handling.

Fixes https://github.com/sonata-project/SonataAdminBundle/issues/4147

## Changelog

```markdown

### Fixed
- Fixed: Checkboxes were not converted using 'iCheck' after Ajax-Calls
```

## To do

- [ ] Update the tests
- [ ] Update the documentation
- [ ] Add an upgrade note

### DIFF
--- a/Resources/views/CRUD/edit_phpcr_one_association_script.html.twig
+++ b/Resources/views/CRUD/edit_phpcr_one_association_script.html.twig
@@ -48,8 +48,8 @@ This code manage the one-to-many association field popup
                     jQuery(form).attr('encoding', 'multipart/form-data');
                 }
                 jQuery('#sonata-ba-field-container-{{ id }}').trigger('sonata.add_element');
-                jQuery('#field_container_{{ id }}').trigger('sonata.add_element');
-                Admin.setup_icheck(jQuery('#field_container_{{ id }}'));
+                var $field_container = jQuery('#field_container_{{ id }}').trigger('sonata.add_element');
+                Admin.setup_icheck($field_container);
             }
         });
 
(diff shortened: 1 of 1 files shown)


Resources/views/CRUD/edit_phpcr_one_association_script.html.twig
```diff
@@ -49,6 +49,7 @@ This code manage the one-to-many association field popup
                 }
                 jQuery('#sonata-ba-field-container-{{ id }}').trigger('sonata.add_element');
                 jQuery('#field_container_{{ id }}').trigger('sonata.add_element');
+                Admin.setup_icheck(jQuery('#field_container_{{ id }}'));
```
greg0ire: `jQuery('#field_container_{{ id }}')` is already computed at the previous line, maybe reuse it?

syeo66: Yes, I was thinking about that after I submitted the pull request. Maybe it would be even better to listen to the sonata.add_element event and do the reformatting stuff (iCheck and maybe Select2 also?) in a more generic way. Wasn't looking into that because I need a quick fix for the iCheck now

greg0ire: > it probably needs to be recomputed anyway

Oh yeah maybe. If it is the case, please add a comment explaining this in the code.

syeo66: No, I was wrong. I was looking 6 lines too high up.

greg0ire:
@syeo66 : next time remove the todo list if you don't intend to actually do it ;)

OskarStark:
I removed the todo list and reworded the changelog message a bit!

Thanks for your contribution @syeo66 👍

PR #414, opened 2017-01-28
Get tree working again with the recent Resource updates

wouterj:
## Changelog

```markdown
### Changed

- Changed data URLs to be in sync with new format returned by the CmfResourceRestbundle
```

## Subject

The RC version of ResourceRestBundle has a slightly modified return format. That causes the actions to stop working. This PR fixes that by adding support for the new format.

*This should be a no-brainer, please merge quickly so we can continue releasing RC and stable versions of the CMF*


greg0ire:
Thanks @wouterj !

PR #421, opened 2017-02-06
1.x to master

dbu:
so there was a conflict. trying to solve it. correct like this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ language: php
 php:
   - '5.6'
   - '7.0'
+  - '7.1'
   - nightly
   - hhvm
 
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -33,29 +34,31 @@ env:
...
+    - php: '7.1'
       env: TARGET=lint
     - php: '5.6'
       env: COMPOSER_FLAGS="--prefer-lowest"
-    - php: '7.0'
+    - php: '7.1'
       env: SYMFONY=2.8.*
-    - php: '7.0'
+    - php: '7.1'
+      env: SYMFONY=3.0.*
```
dbu: or did we not want to test this no longer maintained version? but then composer.json should probably also say ^2.8||^3.1

greg0ire: Does not look like it : https://github.com/sonata-project/dev-kit/blob/master/config/projects.yml

greg0ire: Wait, how come [7.0 is present in the dev-kit config for 1.x](https://github.com/sonata-project/dev-kit/blob/master/config/projects.yml#L183), [but missing from the generated (?) travis config](https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/blob/1.x/.travis.yml)?

greg0ire: [This dev-kit PR](https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/pull/415/files) changed all zeroes into ones…

dbu: which makes sense i would say. just surprising it did no thappen on master too. so can i merge this and the next devkit update will sort things out? merging this seems about right, and will get rid of the merge conflict.

greg0ire: I don't see how this makes sense. 7.0 is still present in the dev-kit config, why would it be removed?

But I think you can safely merge. Whatever the reason for this it, we can figure it out later and keep moving.

dbu: php 7.0 is still in the list of php versions at the top. but i think sonata correctly does not run every permutation of configuration x every php version. rather have one build per php version + one build with the latest php for every configuration.

greg0ire: Still, there should be at least one build with 7.0, don't you think?

dbu: there is, see line 14

dbu: the first 3 lines in the build matrix come from the php versions at the top of the travis config file: https://travis-ci.org/sonata-project/SonataDoctrinePhpcrAdminBundle/builds/198934854

greg0ire: Oh great, I missed that. Everything is ok then.

PR #424, opened 2017-02-09
Filter lower case

### LINKED ISSUE #161: Make filter inside list function case insensitive
Hi,

currently filtering content, as part of list action, is case sensitive. Is it possible to make it insensitive?

thx

ElectricMaxxx:
I am targetting this branch, because *it is for the new release*

Closes #161 

## Changelog

```markdown
### Added
- added `compare_lower_case` Option to `Sonata\DoctrinePHPCRAdminBundle\Filter\StringFilter` to enable lower case comparision
```

## To do

* [x] Update the tests
* [x] Update the documentation (add notice/hint)

## Subject

As described in the issue, we what to enable lowercase comparison on string filter.

### DIFF
--- a/Filter/StringFilter.php
+++ b/Filter/StringFilter.php
@@ -25,28 +25,39 @@ class StringFilter extends Filter
             return;
         }
 
-        $data['value'] = trim($data['value']);
+        $value = trim($data['value']);
         $data['type'] = empty($data['type']) ? ChoiceType::TYPE_CONTAINS : $data['type'];
 
-        if (strlen($data['value']) == 0) {
+        if (strlen($value) == 0) {
             return;
         }
 
         $where = $this->getWhere($proxyQuery);
-
+        $isComparisonLowerCase = $this->getOption('compare_case_insensitive');
+        $value = $isComparisonLowerCase ? strtolower($value) : $value;
         switch ($data['type']) {
             case ChoiceType::TYPE_EQUAL:
-                $where->eq()->field('a.'.$field)->literal($data['value']);
+                if ($isComparisonLowerCase) {
+                    $where->eq()->lowerCase()->field('a.'.$field)->end()->literal($value);
+                } else {
+                    $where->eq()->field('a.'.$field)->literal($value);
+                }
+
                 break;
             case ChoiceType::TYPE_NOT_CONTAINS:
-                $where->fullTextSearch('a.'.$field, '* -'.$data['value']);
+                $where->fullTextSearch('a.'.$field, '* -'.$value);
                 break;
             case ChoiceType::TYPE_CONTAINS:
-                $where->like()->field('a.'.$field)->literal('%'.$data['value'].'%');
+                if ($isComparisonLowerCase) {
+                    $where->like()->lowerCase()->field('a.'.$field)->end()->literal('%'.$value.'%');
+                } else {
+                    $where->like()->field('a.'.$field)->literal('%'.$value.'%');
+                }
+
                 break;
             case ChoiceType::TYPE_CONTAINS_WORDS:
             default:
-                $where->fullTextSearch('a.'.$field, $data['value']);
+                $where->fullTextSearch('a.'.$field, $value);
         }
 
         // filter is active as we have now modified the query
@@ -60,6 +71,7 @@ class StringFilter extends Filter
     {
         return array(
             'format' => '%%%s%%',
+            'compare_lower_case' => false,
         );
     }
 
--- a/Resources/doc/reference/filter_field_definition.rst
+++ b/Resources/doc/reference/filter_field_definition.rst
@@ -179,3 +179,32 @@ implement this functionality.
             return true;
         }
     }
+
+Filtering Fields and Case Sensitivity
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The default behaviour when filtering is to compare values in a case sensitive manner.
+For example "Test" is not the same as "test". Depending on your use case, you might want case insensitive filtering.
+To make the filter case insensitive, use the ``compare_case_insensitive`` option for the string filter:
+
+.. code-block:: php
+
+    <?php
+    namespace My\AppBundle\Admin;
+
+    use Sonata\AdminBundle\Admin\Admin;
+    use Sonata\AdminBundle\Datagrid\DatagridMapper;
+
+    use Application\Sonata\NewsBundle\Document\Comment;
+
+    class PostAdmin extends Admin
+    {
+        protected function configureDatagridFilters(DatagridMapper $datagridMapper)
+        {
+            $datagridMapper
+                ->add('title', 'doctrine_phpcr_string', array('compare_case_insensitive' => true))
+                ->add('author', 'doctrine_phpcr_string', array('compare_case_insensitive' => true))
+                ->add('label', 'doctrine_phpcr_string')
+            ;
+        }
+    }
--- a/Tests/Filter/StringFilterTest.php
+++ b/Tests/Filter/StringFilterTest.php
@@ -16,6 +16,11 @@ use Sonata\DoctrinePHPCRAdminBundle\Form\Type\Filter\ChoiceType;
 
 class StringFilterTest extends BaseTestCase
 {
+    /**
+     * @var StringFilter
+     */
+    private $filter;
+
     public function setUp()
     {
         parent::setUp();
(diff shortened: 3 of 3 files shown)


Resources/doc/reference/filter_field_definition.rst
```diff
@@ -179,3 +179,31 @@ implement this functionality.
             return true;
         }
     }
+
+Filtering case insensitive strings
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Normally the query builder behind the filtering would compare field values case sensitive. There can be reasons to
+filter case insensitive. To do so you can use ``compare_lower_case`` option to force a comparision on field value and
```
greg0ire: - The compare lower case option
- comparison

Resources/doc/reference/filter_field_definition.rst
```diff
@@ -179,3 +179,31 @@ implement this functionality.
...
+filter value
```
greg0ire: Missing final dot

Tests/Filter/StringFilterTest.php
```diff
@@ -16,6 +16,11 @@
 
 class StringFilterTest extends BaseTestCase
 {
+    /**
+     * @var StringFilter
+     */
+    protected $filter;
```
greg0ire: Private?

ElectricMaxxx:
Reading you hints on the docs points me on a question:

using lower case is the solution. the solution for the problem of case insensitive filtering. Is it correct to show and document the solution or should talk (even in the option) about casse insensitive comparision?

greg0ire:
Making it abstract might be better indeed, at least in the name of the option. You can still document the solution by making a separate §, starting with "under the hood we do this…". Also, if this was SQL, lowercase would not be the best solution, because it prevents from using indexes. Maybe there is the same problem in phpcr, I don't know.

ElectricMaxxx:
Might be the same in our dbal implementation, but i won't see it that hard.Changed the open stuff and renamed the option.

Resources/doc/reference/filter_field_definition.rst
```diff
@@ -184,7 +184,7 @@ Filtering case insensitive strings
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Normally the query builder behind the filtering would compare field values case sensitive. There can be reasons to
-filter case insensitive. To do so you can use ``compare_lower_case`` option to force a comparision on field value and
+filter case insensitive. To do so you can use ``compare_case_insensitive`` option to force a comparision on field value and
```
greg0ire: It's "comparison" :P

Resources/doc/reference/filter_field_definition.rst
```diff
@@ -179,3 +179,31 @@ implement this functionality.
             return true;
         }
     }
+
+Filtering case insensitive strings
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Normally the query builder behind the filtering would compare field values case sensitive. There can be reasons to
+filter case insensitive. To do so you can use ``compare_case_insensitive`` option to force a case insensitive
```
greg0ire: "The" blah blah blah option

greg0ire:
> Might be the same in our dbal implementation

I don't recall seeing anything like that, probably b/c it works by default with mysql. For postgresql , you can do [this](http://stackoverflow.com/questions/16247684/apply-custom-doctrine-output-walker-by-default-to-all-queries)

ElectricMaxxx:
You can find the QB documentation here: http://docs.doctrine-project.org/projects/doctrine-phpcr-odm/en/latest/reference/query-builder.html

generally i have no clue what happens between the the PHPCR query and dbal as an implementation, but i trust the query builder :-)

Resources/doc/reference/filter_field_definition.rst
```diff
@@ -179,3 +179,31 @@ implement this functionality.
             return true;
         }
     }
+
+Filtering case insensitive strings
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Normally the query builder behind the filtering would compare field values case sensitive. There can be reasons to
+filter case insensitive. To do so you can use The ``compare_case_insensitive`` option to force a case insensitive
```
greg0ire: I was on my phone when I wrote this and omitted to put "the" in lowercase, I thought you would naturally fix it :P

ElectricMaxxx: There is nothing natural whith my grammar. Even in my home language. ;-)

Filter/StringFilter.php
```diff
@@ -25,28 +25,39 @@ public function filter(ProxyQueryInterface $proxyQuery, $alias, $field, $data)
...
+                if ($isComparisonLowerCase) {
+                    $where->eq()->lowerCase()->field('a.'.$field)->end()->literal($value);
+                } else {
+                    $where->eq()->field('a.'.$field)->literal($value);
+                }
+
                 break;
             case ChoiceType::TYPE_NOT_CONTAINS:
-                $where->fullTextSearch('a.'.$field, '* -'.$data['value']);
+                $where->fullTextSearch('a.'.$field, '* -'.$value);
```
ElectricMaxxx: this is old, wasn't submitted only

ElectricMaxxx:
@dbu @greg0ire time to merge that one?

greg0ire:
I don't merge unless there is already one approval other than mine, sorry.

OskarStark:
I don't understand why we don't use `1.x` branch, it looks BC to me

ElectricMaxxx:
@OskarStark as the behavior defaults to ther prior behavior, this is no BC. It is just a new feature available on 2.x then.

greg0ire:
> as the behavior defaults to ther prior behavior

This looks kinda BC when you describe it like that…

Filter/StringFilter.php
```diff
@@ -25,28 +25,39 @@ public function filter(ProxyQueryInterface $proxyQuery, $alias, $field, $data)
             return;
         }
 
-        $data['value'] = trim($data['value']);
+        $value = $data['value'] = trim($data['value']);
```
dbu: why update $data['value'] here? you never read from it below. i am ok moving this to $value

Resources/doc/reference/filter_field_definition.rst
```diff
@@ -179,3 +179,31 @@ implement this functionality.
             return true;
         }
     }
+
+Filtering case insensitive strings
```
dbu: i would call the title "Filtering Fields and Case Sensitivity"

Resources/doc/reference/filter_field_definition.rst
```diff
@@ -179,3 +179,31 @@ implement this functionality.
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Normally the query builder behind the filtering would compare field values case sensitive. There can be reasons to
+filter case insensitive. To do so you can use the ``compare_case_insensitive`` option to force a case insensitive
+comparison on field value and filter value.
```
dbu: maybe like this?

The default behaviour when filtering is to compare values in a case sensitive manner. For example "Test" is not the same as "test". Depending on your use case, you might want case insensitive filtering. To make the filter case insensitive, use the ``compare_case_insensitive`` option for the string filter:

Resources/doc/reference/filter_field_definition.rst
```diff
@@ -179,3 +179,31 @@ implement this functionality.
...
+    use Application\Sonata\NewsBundle\Document\Comment;
+
+    class PostAdmin extends Admin
+    {
+        protected function configureDatagridFilters(DatagridMapper $datagridMapper)
+        {
+            $datagridMapper
+                ->add('title')
+                ->add('author')
+                ->add('with_case_insensitive_comparison', 'doctrine_phpcr_string', array('compare_case_insensitive' => true))
```
dbu: i would make the title and author fields case insensitive filtering to make the example more real life

Resources/doc/reference/filter_field_definition.rst
```diff
@@ -179,3 +179,31 @@ implement this functionality.
...
+            ;
+        }
```
dbu: can you close the last bracket please?

Filter/StringFilter.php
ElectricMaxxx: didn't wanna change the former behavior. Do not know if somebody use the trimmed value afterwards and expects it to be trimmed.

ElectricMaxxx: cause you can change the filters, right? I don't ever use the array value again, as i use a local var.

Resources/doc/reference/filter_field_definition.rst
ElectricMaxxx: I used a label field now to also show a case sensitive filtering

Filter/StringFilter.php
ElectricMaxxx: but, you are right. Depending on a magic behavior in a chain item before isn't a really good approach.

dbu: ok, valid point. but then i'd just keep using $data['value'] below for simplicity. having a local variable does not add much

Resources/doc/reference/filter_field_definition.rst
```diff
@@ -179,3 +179,32 @@ implement this functionality.
             return true;
         }
     }
+
+Filtering Fields and Case Sensitivity
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
dbu: this needs to be aligned with the title. its the reason the doc build fails

ElectricMaxxx:
@greg0ire should i sqaush again then?

greg0ire:
You SHOULD (RFC 2119)

ElectricMaxxx:
@greg0ire i really DID so.

greg0ire:
I'll merge when Travis is green.

greg0ire:
Cancelling old builds to speed things up

dbu:
great job, thanks max and thanks gregoire for reviewing!

PR #426, opened 2017-02-12
Fixed inline editing by removing assets from the form type

### LINKED ISSUE #327: A better way of importing JS resources for tree input
Currently, the JS and CSS files required for the JS input are loaded at the field itself. This doesn't result in a very great page rendering:
1. The page renders until the tree field
2. It then waits to load all the files and execute the javascript
3. After that happend, the rest of the page is rendered

It also means that including the tree field 3 times would lead to loading all assets 3 times.

wouterj:
Closes #327

## Changelog

```markdown
### Added
- Added `AddTreeBrowserAssetsPass` to load TreeBrowser assets using the admin pool
- Added a custom form theme for the TreeSelectType in order to use AdminLTE
```

## Subject

When adding support for inline editing, an AJAX request is done to load the form. Since the TreeSelectType includes `<script>` tags, calling `.replaceWith(html)` will result in a blank page. Also, as you can read in #327, loading fancytree for each usage of the form type isn't needed. Loading it only once in the head would be much better.

This PR fixes both issues by configuring the admin pool to include the fancytree assets. It also overrides the TreeSelectType to avoid loading these assets again.

Meanwhile, I also added some AdminLTE styles to the TreeSelectType to make styling more consistent and nicer. 

| **Before** | **After**
| --- | ---
| cmf-tree-style-before | cmf-tree-style-after

### DIFF
--- a/DependencyInjection/Compiler/AddTreeBrowserAssetsPass.php
+++ b/DependencyInjection/Compiler/AddTreeBrowserAssetsPass.php
@@ -21,7 +21,7 @@ use Symfony\Component\DependencyInjection\Reference;
  *
  * @author Wouter de Jong <wouter@wouterj.nl>
  */
-class AddTreeBrowserAssetsPass implements CompilerPassInterface
+final class AddTreeBrowserAssetsPass implements CompilerPassInterface
 {
     /**
      * {@inheritdoc}
@@ -32,9 +32,7 @@ class AddTreeBrowserAssetsPass implements CompilerPassInterface
             return;
         }
 
-        $definition = $container->findDefinition('sonata.admin.pool');
-        $this->addAssetsToAdminPool($definition);
-
+        $this->addAssetsToAdminPool($container->findDefinition('sonata.admin.pool'));
         $this->addFormResources($container);
     }
 
--- a/DependencyInjection/Compiler/AddTreeBrowserAssetsPass.php
+++ b/DependencyInjection/Compiler/AddTreeBrowserAssetsPass.php
@@ -1,0 +1,57 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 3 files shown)


DependencyInjection/Compiler/AddTreeBrowserAssetsPass.php
```diff
@@ -0,0 +1,57 @@
...
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('sonata.admin.pool')) {
+            return;
+        }
+
+        $definition = $container->findDefinition('sonata.admin.pool');
+        $this->addAssetsToAdminPool($definition);
```
greg0ire: Please remove the one-time variable : `$this->addAssetsToAdminPool($container->findDefinition('sonata.admin.pool'));`

Resources/views/Form/tree_browser_fields.html.twig
```diff
@@ -0,0 +1,38 @@
...
+    </div>
+{% endblock %}
+
+{% block cmf_tree_select_widget_input %}
+    <div class="input-group">
+        {{ block('form_widget_simple') }}
+
+        {% if "compact" == widget %}
+            <div class="input-group-btn">
+                <button class="btn btn-primary" id="{{ id }}-cmf-tree-tooltip-toggle">{{ 'form.button.pick'|trans({}, 'CmfTreeBrowserBundle') }}</button>
```
greg0ire: Please linebreak this.

DependencyInjection/Compiler/AddTreeBrowserAssetsPass.php
```diff
@@ -0,0 +1,57 @@
...
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Adds javascripts and stylesheets required for the TreeSelectType.
+ *
+ * @author Wouter de Jong <wouter@wouterj.nl>
+ */
+class AddTreeBrowserAssetsPass implements CompilerPassInterface
```
greg0ire: Could be final

wouterj:
Thanks for the quick review each time @greg0ire! Should be fixed now.

greg0ire:
Thanks @wouterj !

PR #435, opened 2017-04-23
Move and reorder and sorable tree

### LINKED ISSUE #434: Fix treebrowser options in view
Depending on discussion on https://github.com/symfony-cmf/tree-browser-bundle/issues/146 we have to update the option for the treebrowser to enable dnd again.

ElectricMaxxx:
I am targeting this branch, because it should get into 2.0@dev

Closes #434 
fix #434, https://github.com/symfony-cmf/tree-browser-bundle/issues/146, https://github.com/symfony-cmf/tree-browser-bundle/issues/139

## Changelog

```markdown
### Added
- Added options (implementation of move) passed to the tree, to enable drag&drop. 
```

## Subject

We had problems to see drag&drop working on the tree, now it should work again. treebrowser needed an implementation of the `move` on the tree repository. This is done by two ajax requests now.

## ToDo

* [x] move here and in tree-browser
* [x] reorder in Backend
* [x] reorder works in tree browser
* [ ] answers on questions

### DIFF
--- a/Controller/TreeController.php
+++ b/Controller/TreeController.php
@@ -14,6 +14,7 @@ namespace Sonata\DoctrinePHPCRAdminBundle\Controller;
 use Doctrine\Bundle\PHPCRBundle\ManagerRegistry;
 use PHPCR\Util\PathHelper;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 
@@ -111,12 +112,16 @@ class TreeController extends Controller
      */
     public function reorderAction(Request $request)
     {
-        $parent = $request->request->get('parent');
-        $moved = $request->request->get('dropped');
-        $target = $request->request->get('target');
-        $position = $request->request->get('position');
-        $before = 'before' == $position;
+        $parent = $request->get('parent');
+        $moved = $request->get('dropped');
+        $target = $request->get('target');
+        $position = $request->get('position');
 
+        if (null === $parent || null === $moved || null === $target) {
+            return new JsonResponse(['Parameters parent, dropped and target has to be set to reorder.'], Response::HTTP_BAD_REQUEST);
+        }
+
+        $before = 'before' == $position;
         $parentNode = $this->session->getNode($parent);
         $targetName = PathHelper::getNodeName($target);
         if (!$before) {
--- a/Controller/TreeController.php
+++ b/Controller/TreeController.php
@@ -11,6 +11,8 @@
 
 namespace Sonata\DoctrinePHPCRAdminBundle\Controller;
 
+use Doctrine\Bundle\PHPCRBundle\ManagerRegistry;
+use PHPCR\Util\PathHelper;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -41,13 +43,26 @@ class TreeController extends Controller
     private $confirmMove = false;
 
     /**
+     * @var \PHPCR\SessionInterface
+     */
+    private $session;
+
+    /**
+     * @param ManagerRegistry $manager
+     * @param string $sessionName
      * @param string $repositoryName
      * @param string $template
-     * @param array  $defaults
-     * @param bool   $confirmMove
+     * @param array $defaults
+     * @param bool $confirmMove
      */
-    public function __construct($repositoryName = 'default', $template = null, array $defaults = array(), $confirmMove = false)
-    {
+    public function __construct(
+        ManagerRegistry $manager,
+        $sessionName,
+        $repositoryName = 'default',
+        $template = null,
+        array $defaults = array(),
+        $confirmMove = false
+    ) {
         $this->repositoryName = $repositoryName;
         if ($template) {
             $this->template = $template;
@@ -55,6 +70,8 @@ class TreeController extends Controller
         $this->defaults = $defaults;
 
         $this->confirmMove = $confirmMove;
+
+        $this->session = $manager->getConnection($sessionName);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


ElectricMaxxx:
As the process is confusing again, i need to add the upgrad note on 1.x right?

ElectricMaxxx:
@dbu @wouterj can you have a look at the change. Tested it in sandbox and it works fine now.

greg0ire:
> Add an upgrade note ???

Is there a BC-break? If yes, there should be an upgrade note, filed under the paragraph corresponding to the version the BC-break will come with.

ElectricMaxxx:
I don't think that this is a BC-Break as we slowly re-enable an feature that worked on 1.x. But unfortunately the positioning still does not work. I can move, but the resource-api is able to move and append only. We would need an patch on the position property of the node too. I would also keep that implementation in here and pass a position  from the treebrowser only, so the tree implementation doesn't need to know about the resource api again

Controller/TreeController.php
```diff
@@ -41,20 +43,35 @@ class TreeController extends Controller
...
+
+    /**
+     * @param ManagerRegistry $manager
+     * @param string $sessionName
      * @param string $repositoryName
      * @param string $template
-     * @param array  $defaults
-     * @param bool   $confirmMove
+     * @param array $defaults
+     * @param bool $confirmMove
```
dbu: we seem to not use styleci on this repo, but can you please still align the variables?

Resources/config/tree.xml
```diff
@@ -2,6 +2,8 @@
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
         <service id="sonata.admin.doctrine_phpcr.tree_controller" class="Sonata\DoctrinePHPCRAdminBundle\Controller\TreeController">
+            <argument type="service" id="doctrine_phpcr" />
```
dbu: the linter complains about the whitespace before />

ElectricMaxxx:
shall we mark (changelog, code) somethin as deprecated then, or is the
implementation something hidden, that no-one would touch?

Btw: I still try to get it working in sandbox, currently the controller
isn't instantiated correctly (was it used before?). And we need to add
the usage of the `reorder()` in the tree-browser js code, but i got a
draft for that almost ready too.

dbu:
> deprecated

no, but we could mark it as internal. but even then, people need to copy the twig template if they want to customize something, so the method will have to stay during the major version. i don't see a problem with having it stay.

ElectricMaxxx:
I think i have the reorder working from this bundle, what i miss is the ability of the tree to handle the position and enrich the resource with the position value.

Controller/TreeController.php
```diff
@@ -84,4 +102,48 @@ public function treeAction(Request $request)
...
+            }
+            $targetName = null;
+            if ($nodesIterator->valid()) {
+                $nodesIterator->next();
+                if ($nodesIterator->valid()) {
+                    $targetName = $nodesIterator->key();
+                }
+            }
+        }
+        $parentNode->orderBefore($targetName, PathHelper::getNodeName($dropedAtPath));
```
ElectricMaxxx: @dbu  this one was in the wrong direction on `1.x` (where i copied the code), so that can't have ever worked.

Description/PositionEnhancer.php
```diff
@@ -0,0 +1,79 @@
...
+        $counter = 0;
+        while ($nodeIterator->valid()) {
+            $counter++;
+            if ($nodeIterator->key() === $nodeName) {
+                break;
+            }
+            $nodeIterator->next();
+        }
+
+        $description->set('position', $counter);
```
ElectricMaxxx: @wouterj @dbu is that name ok?

Description/PositionEnhancer.php
```diff
@@ -0,0 +1,79 @@
...
+            return false;
+        }
+
+        try {
+            $parentNode = $this->session->getNode(PathHelper::getParentPath($resource->getPath()));
+        } catch (PathNotFoundException $exception) {
+            return false;
+        }
+
+        // Todo: check for non orderable type
```
ElectricMaxxx: @dbu @lsmith77 @dantleech  do we have that orderable Type on a phpcr node? (https://docs.adobe.com/content/docs/en/spec/jcr/2.0/23_Orderable_Child_Nodes.html)

Resources/views/Tree/tree.html.twig
```diff
@@ -46,7 +88,9 @@ file that was distributed with this source code.
                     label: "{{ 'delete_item'|trans({}, 'SonataDoctrinePhpcrAdminBundle') }}",
                     icon: 'fa fa-trash-o'
                 }
-            }
+            },
+            dnd: {enabled: true, reorder: true},
+            sortableBy: 'position'
```
ElectricMaxxx: @wouterj  is it ok to have a single option for that on tree implementation? or just into dnd section?

ElectricMaxxx:
Argh, ... isn't ready completely, still got a problem to sort on tree. Backend is fine.

ElectricMaxxx:
@symfony-cmf/contributers Works now, ready to revierw. Working example on:
https://github.com/symfony-cmf/cmf-sandbox/pull/366

Controller/TreeController.php
```diff
@@ -84,4 +102,48 @@ public function treeAction(Request $request)
...
+        $dropedAtPath = $request->get('dropped');
+        $targetPath = $request->get('target');
+        $position = $request->get('position');
+
+        if (null === $parentPath || null === $dropedAtPath || null === $targetPath) {
+            return new JsonResponse(['Parameters parent, dropped and target has to be set to reorder.'], Response::HTTP_BAD_REQUEST);
+        }
+
+        $before = 'before' == $position;
+        $parentNode = $this->session->getNode($parentPath);
```
pbeliberty: How is set the repository name please ? 
In the [resource controller](https://github.com/symfony-cmf/resource-rest-bundle/blob/master/src/Controller/ResourceController.php#L79) there is 
```
            $repository = $this->registry->get($repositoryName);

            $fullPath = method_exists($repository, 'resolvePath') ? $repository->resolvePath($path) : $path;
```

Controller/TreeController.php
OskarStark: indeed, we should use StyleCI here, @Soullivaneuh can you support us in setting this up?

Description/PositionEnhancer.php
```diff
@@ -0,0 +1,79 @@
...
+ * @author Maximilian Berghoff <Maximilian.Berghoff@mayflower.de>
+ */
+class PositionEnhancer implements DescriptionEnhancerInterface
+{
+    /**
+     * @var \PHPCR\SessionInterface
+     */
+    private $session;
+
+    public function __construct(ManagerRegistry $manager, $sessionName)
```
OskarStark: phpdoc missing

Controller/TreeController.php
ElectricMaxxx: The repositories in resource component are a different. In that case they are coupled to the resources. In case of sonata we do not have them.

ElectricMaxxx: but we can/should go through the model manager implemented as facedes in sonata environments, isn't it?

ElectricMaxxx:
Looks good now. Tests are (almost) green. 

I created a  new issue to make the configuration great again #441

ElectricMaxxx:
I merge to verify this last feature on our sandbox. Should work, bug tickets else.

OskarStark:
@ElectricMaxxx please squash your commits in the future, I don't think we need 40 commits:
https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/commits/master

And please don't merge your own PR's (except some patches), thank you!

Controller/TreeController.php
dbu: i am now enabling styleci and will let it make the code according to the dev-kit styleci.yml configuration.

Issue #440, opened 2017-06-07
lowest version build times out


dbu:
the lowest version build on master times out. we should see about restricting the options or something so that it does not time out anymore

greg0ire:
Related : https://github.com/sonata-project/SonataCommentBundle/pull/68#issuecomment-305969242

greg0ire:
Fixed by not having a gazillion branches in the deps of this package.

PR #446, opened 2017-06-30
trying to fix lowest-version build

dbu:
on master, the lowest-version build times out (#440). trying to see if restricting the dependencies a bit helps composer find a solution.

this is not mergeable because .travis/ is managed by dev-kit...

### DIFF
--- a/.travis/install_test.sh
+++ b/.travis/install_test.sh
@@ -18,6 +18,7 @@ chmod u+x "${HOME}/bin/coveralls"
 
 # To be removed when this issue will be resolved: https://github.com/composer/composer/issues/5355
 if [ "${COMPOSER_FLAGS}" = '--prefer-lowest' ]; then
-    composer update --prefer-dist --no-interaction --prefer-stable --quiet
+    composer require --no-install symfony/framework-bundle 2.8.*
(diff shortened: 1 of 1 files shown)


dbu:
@greg0ire so this change would fix the problem for us. but i guess the file is managed by dev-kit, so its not really a solution. any idea how we can achieve this with dev-kit?

greg0ire:
First, let's see how #450 goes, shall we?

greg0ire:
@dbu it seems to be working. Do you like my solution?

dbu:
the problem is only with master, not with the 1.x branch. i merged your pr to 1.x but now we need the same for master.

Soullivaneuh:
This is why I added [`travis_wait`](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received) before on dev-kit.

This was removed because "no build is under timeout now" but I think this should be restored, by prevention.

@dbu Could you try with this method on this PR?

dbu:
lets see if waiting is enough

dbu:
yay! see https://github.com/sonata-project/dev-kit/pull/252

PR #450, opened 2017-07-03
Bump every dep to its latest patch version

greg0ire:
This will reduce the space of possible versions and hopefully, fix the
prefer-lowest build.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,13 +17,13 @@
     ],
     "require": {
         "php": "^5.3.3 || ^7.0",
-        "doctrine/phpcr-bundle": "^1.1",
-        "doctrine/phpcr-odm": "^1.1",
+        "doctrine/phpcr-bundle": "^1.1.2",
+        "doctrine/phpcr-odm": "^1.1.4",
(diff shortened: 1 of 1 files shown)


dbu:
this should normally be fine. though it might trip up some people on edge cases, where they pin to a lower patch version because of some obscure incompatibility between dependencies. (every bugfix is a small BC break somehow)

greg0ire:
> this should normally be fine. though it might trip up some people on edge cases, where they pin to a lower patch version because of some obscure incompatibility between dependencies. (every bugfix is a small BC break somehow)

Yeah not great, but should be fine indeed.

@lsmith77 and @Soullivaneuh 's bots are fighting over labels lol

dbu:
@lsmith77 can you remove your bot from the sonata project please?

can you do the pull request from 1.x to master? there will be conflicts because composer.json looks quite a bit different... i did a branch from master, merged 1.x into it and solved the conflicts and then did a PR with that against master.

greg0ire:
Oh... so maybe my fix won't fix anything after all...

greg0ire:
See #451

PR #451, opened 2017-07-03
Merge 1.x

greg0ire:
Manual merge. Hopefully fixes timeout issue.

PR #452, opened 2017-07-03
Revert "Bump every dep to its latest patch version"

greg0ire:
Reverts sonata-project/SonataDoctrinePhpcrAdminBundle#450

Wasn't useful after all, see #451


dbu:
thanks. i guess i should not have merged the first one - we could have done this directly against master only, as nothing was wrong with 1.x

PR #453, opened 2017-07-04
Clear configuration

### LINKED ISSUE #441: Configuration for treebrowser options
## Subjecton

We have some configuration that isn't used any more and we do have new options on the new treebrowser which should be definded as bundle configuration again.

## List of existing configurations

- document_tree_options.confirm_move
- document_tree_options.precise_children 
- document_tree_config. * (no usage anymore atm)
- document_tree_defaults (is more a routing default configuration)

### New options on treebrowser

- sortableBy
- drag & drop
-- enabled
-- reorder

ElectricMaxxx:
Closes #441 fix #441 

## Changelog
```markdown
### Added

- enclosed `document_tree` configuration including options for the tree-browser" to "completely changed `document_tree` configuration to work with the CmfResourceBundle. Tree options are now directly under `document_tree`.

### Removed

- old `document_tree` configuration
```

## To do

- [x] Update the tests
- [x] Update the documentation

## Subject

This is a way i would suggest to refactor the configuration to also pass some options into the tree-browser.


ElectricMaxxx:
Question @greg0ire How shall i deprecate the configuration on `1.x`? Upgrade notes only?

ElectricMaxxx:
@SonataCI done :-)

DependencyInjection/Configuration.php
```diff
@@ -65,41 +64,27 @@ public function getConfigTreeBuilder()
...
                     ->children()
-                        ->integerNode('depth')
-                            ->defaultValue(1)
-                            ->info('Depth to which to fetch tree children when rendering the initial tree')
+                        ->scalarNode('repository_name')
+                            ->defaultValue('default')
+                            ->info('The repository name the resource API connects to.')
+                        ->end()
+                        ->arrayNode('routing_defaults')
+                            ->prototype('scalar')->end()
```
dbu: can you add an info node for this too please?

DependencyInjection/Configuration.php
```diff
@@ -65,41 +64,27 @@ public function getConfigTreeBuilder()
...
                         ->end()
-                        ->scalarNode('precise_children')
-                            ->defaultTrue()
-                            ->info('Exact check if document has children. For large trees, set to false for better performance, but user might needs to click on expand to see there are no children.')
+                        ->scalarNode('sortable_by')
+                            ->defaultValue('position')
+                            ->info('Defines the document property siblings are sorted.')
```
dbu: Defines by which property to sort sibling documents.

DependencyInjection/Configuration.php
```diff
@@ -65,41 +64,27 @@ public function getConfigTreeBuilder()
...
                         ->end()
-                        ->booleanNode('confirm_move')
-                            ->defaultTrue()
-                            ->info('Whether moving a node in the tree asks for confirmation.')
+                        ->arrayNode('move')
+                            ->canBeEnabled()
+                            ->children()
+                                ->booleanNode('reorder')
+                                    ->defaultTrue()
+                                    ->info('Ordering after a move action is enabled, causes new requests.')
```
dbu: lets leave the explanations for the documentation and keep the info nodes minimal. "Whether to enable reordering sibling documents."

btw, i am not 100% sure this should be optional. i feel like if move is enabled, reordering should implicitly be allowed. what is the reasoning to make it optional?

Resources/views/Tree/tree.html.twig
```diff
@@ -10,7 +10,7 @@ file that was distributed with this source code.
 #}
 
 {% set routing_default_values = {} %}
-{% for default in routing_defaults %}
+{% for default in tree_configuration['routing_defaults'] %}
```
dbu: afaik this could also be written as tree_configuration.routing_defaults (same for the fields below)

DependencyInjection/Configuration.php
ElectricMaxxx: yea, that is the option i am struggeling with too.

I implemented it on the tree-browser to have 2 requests less. Reason is:
The move call on the resource API needs two requests, as move itself answers `204 No Content". And move simply does a move only - means a change on the path. So there are systems, where the position does not matter (routing, content). But for nodes where the position matters, we need to patch the position too (menu stuff). So when `reorder` is enabled a request will patch the position. Not directly by changing the value, cause position does not exist as value. The request triggers an `insertBefore()`

BTW: This need to go into docs now. So the discussion should give the right focus.

Resources/views/Tree/tree.html.twig
ElectricMaxxx: ah, yes, forgot it.

ElectricMaxxx:
@dbu The reason to bundle the configuration into in array is caused by the amount of constructor parameters. I think, we should not add a new parameter for the constructor for every little option we'd like to pass to the tree-browser. We can hold those values as single values and in the collection, but i don't think that we should each option one by one into the constructor.

Controller/TreeController.php
```diff
@@ -26,53 +26,31 @@ class TreeController extends Controller
...
+    /**
+     * @var array
+     */
+    private $treeConfiguration;
 
     /**
      * @param ManagerRegistry $manager
      * @param string          $sessionName
-     * @param string          $repositoryName
+     * @param array           $treeConfiguration
```
dbu: then lets add a docblock here that this is the same structure as defined in the Configuration class, apart from move becoming a boolean.

Controller/TreeController.php
```diff
@@ -85,21 +63,11 @@ public function __construct(
...
 
         return $this->render($this->template, array(
-            'repository_name' => $this->repositoryName,
             'root_node' => $root,
-            'routing_defaults' => $this->defaults,
-            //'confirm_move' => $this->confirmMove,
-            //'create_in_overlay' => $createInOverlay ? $createInOverlay : false,
-            //'edit_in_overlay' => $editInOverlay ? $editInOverlay : false,
-            //'delete_in_overlay' => $deleteInOverlay ? $deleteInOverlay : false,
+            'tree_configuration' => $this->treeConfiguration,
```
dbu: lets explode the array here and pass explicit named parameters to the template as before. then its clear what we expect to exist and the template will be simpler.

DependencyInjection/Configuration.php
dbu: as this is the config for the tree on the dashboard, i think i would just drop that configuration here and set the value to true in the DI extension. if somebody does a custom tree somewhere else its ok if controller and template support disabling the reorder but for the dashboard i don't see a use case where it would make sense to move without reorder. then that whole discussion can remain in the resource bundle doc and the admin doc does not need to bother with the details of tree things.

and i guess some upcoming resource bundle version should support a combined move and reorder operation to go to 1 instead of 3 requests, but that is post 2.0 release...

DependencyInjection/Configuration.php
ElectricMaxxx: Yea complicated. I do not really even know if we still have to keep it in configuration. mostly that are the routing default values where merged with some other defaults (which could be done in the di extension) and passed to the `GET /resource` route.

DependencyInjection/Configuration.php
ElectricMaxxx: And we should add step by step some edit/patch support, to remove the patch route we have in this repository.

ElectricMaxxx:
btw: we still have no real `WebTest` to see the tree working.

ElectricMaxxx:
@dbu  StyleCi tries to force me to horizontal allignment of parameter doc blocks.

ElectricMaxxx:
I still miss the schema file to update it.

dbu:
can you apply styleci again? btw, the styleci template is here https://github.com/sonata-project/dev-kit/blob/master/project/.styleci.yml - we should see with the sonata people if we can activate things like short array notation (but that would require that every single sonata project only uses php versions that support it)

there is no schema for the xml configuration in this repository. your pull request is still marked as [POC] but i think its the right direction. so whats missing is an entry in the UPGRADE-1.x.md file. (though i think that should be called UPGRADE-2.x.md instead and move the 2.0 note that is currently in the 1.x file to the new 2.x file). anything else?

Resources/config/tree.xml
```diff
@@ -4,10 +4,8 @@
         <service id="sonata.admin.doctrine_phpcr.tree_controller" class="Sonata\DoctrinePHPCRAdminBundle\Controller\TreeController">
             <argument type="service" id="doctrine_phpcr"/>
             <argument>%doctrine_phpcr.default_session%</argument>
-            <argument>%sonata_admin_doctrine_phpcr.tree_block.repository_name%</argument>
+            <argument>%sonata_admin_doctrine_phpcr.tree_block.configuration%%</argument>
```
Soullivaneuh: Extra `%` here, not needed I guess.

composer.json
```diff
@@ -29,6 +29,7 @@
     "require-dev": {
         "dantleech/glob-finder": "1.*",
         "jackalope/jackalope-jackrabbit": "^1.0",
+        "matthiasnoback/symfony-dependency-injection-test": "1.1.*",
```
Soullivaneuh: `^1.1`

Resources/config/tree.xml
Soullivaneuh: BTW, I think it's not tested at all her. Travis is currently green.

ElectricMaxxx: I think we should add WebTests at all to test the controllers and internals. So i would suggest: #457

dbu: agreed. but if you tested the changes manually i am ok with merging, as it fixes the existing situation. #457 should be done separately.

DependencyInjection/Configuration.php
```diff
@@ -65,41 +64,19 @@ public function getConfigTreeBuilder()
...
-
-                ->arrayNode('document_tree_options')
-                    ->addDefaultsIfNotSet()
+                    ->canBeEnabled()
                     ->children()
-                        ->integerNode('depth')
-                            ->defaultValue(1)
-                            ->info('Depth to which to fetch tree children when rendering the initial tree')
+                        ->scalarNode('repository_name')
+                            ->defaultValue('default')
```
dbu: should this not be null by default? when we ask the manager-registry for the default repository, we should ask for `null` rather than the one called "default". "default" is the default name for the default repository, but afaik `null` is the correct parameter to ask for the default repo (as the name could be changed to something else)

dbu:
for the "Added" section: can we rephrase "enclosed `document_tree` configuration including options for the tree-browser" to "completely changed `document_tree` configuration to work with the CmfResourceBundle. Tree options are now directly under `document_tree`."

DependencyInjection/Configuration.php
ElectricMaxxx: The tree does not ask the manager registry to fetch the repositor. The tree creates an api url lik `/de/admin/default/cms/content/content-1` So the repository name is a string in this bundle only. 
I am not sure if i am able to set the default name in case of `null` inside the bundles extension class. Or the admin should do it. I will try the last idea first.

ElectricMaxxx:
@dbu I added a `prepend()` to this extension class to set the `repository_name` to the one of `CmfResourceBundle`. Hard to test, but it work now. Do you like it? The only think i miss now, is the confusing update/grade note.

DependencyInjection/Configuration.php
```diff
@@ -65,41 +64,19 @@ public function getConfigTreeBuilder()
...
-                 ->end()
-
-                 ->scalarNode('document_tree_repository')->defaultValue('default')->end()
-
-                ->arrayNode('document_tree_defaults')
-                    ->prototype('scalar')->end()
-                ->end()
-
-                ->arrayNode('document_tree_options')
-                    ->addDefaultsIfNotSet()
```
dbu: i think you need this line, no?

DependencyInjection/SonataDoctrinePHPCRAdminExtension.php
```diff
@@ -83,79 +88,56 @@ public function getNamespace()
...
 
     /**
-     * Set the tree type mapping configuration in the services.
-     *
-     * @param array            $config
-     * @param ContainerBuilder $container
+     * {@inheritdoc}
      */
-    private function loadTreeTypes($config, ContainerBuilder $container)
+    public function prepend(ContainerBuilder $container)
```
dbu: so to prepend configuration, we load our own configuration already and process it, then process the cmf_resource configuration and prepend values from that to our own configuration? this feels super complicated. what exactly are we trying to achieve with this? i think we need to find a more straightforward solution. maybe i have an idea when i understand what exactly we try to do here.

ElectricMaxxx:
When `repository_name` frpm our configuration is null, we have to set the default one of `cmf_resourse`. We can also do the default setting in the controller also, but is that clean?-------- Originalnachricht --------Betreff: Re: [sonata-project/SonataDoctrinePhpcrAdminBundle] Clear configuration (#453)Von: David Buchmann An: sonata-project/SonataDoctrinePhpcrAdminBundle Cc: Maximilian Berghoff ,Author @dbu commented on this pull request.



In DependencyInjection/SonataDoctrinePHPCRAdminExtension.php:
>       */
-    private function loadTreeTypes($config, ContainerBuilder $container)
+    public function prepend(ContainerBuilder $container)

so to prepend configuration, we load our own configuration already and process it, then process the cmf_resource configuration and prepend values from that to our own configuration? this feels super complicated. what exactly are we trying to achieve with this? i think we need to find a more straightforward solution. maybe i have an idea when i understand what exactly we try to do here.

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.


  
  




{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/sonata-project/SonataDoctrinePhpcrAdminBundle","title":"sonata-project/SonataDoctrinePhpcrAdminBundle","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle"}},"updates":{"snippets":[{"icon":"PERSON","message":"@dbu commented on #453"}],"action":{"name":"View Pull Request","url":"https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/pull/453#pullrequestreview-50053658"}}}

ElectricMaxxx:
@dbu you have an idea?

DependencyInjection/SonataDoctrinePHPCRAdminExtension.php
ElectricMaxxx: When `repository_name` for our configuration is null, we have to set the default one of `cmf_resourse`. We can also do the default setting in the controller also, but we can not use `null`at all?

So we have 3 possibilities from my POV:
- fetch the default value from `resource-bundle`
- force the user to set it on its own – would do it twice now
- pass the default value of `resource-bundle` to our controller, and use that in case of a `null`

Do you have an other idea?

dbu:
do we need to fetch the correct manager both in this bundle and in the resource bundle? is there no service from the resource bundle that could give access to the manager, so we could inject that service in the sonata bundle instead?

if non of this is the case, we should have the resource bundle prepend to sonata admin, not sonata prepend the resource config to itself. or could we simply reuse the parameter that the resource bundle sets in the container, if we are not overridden? does it even make sense to have the admin configurable separately from the resource bundle? if both bundles provide services that must be talking to the same admin to make sense, i don't see why we should allow configuring a different admin here.

ElectricMaxxx:
@dbu  i used the easies solution now, i passed the parameter of the `default_name` of `CmfResouceBundle` into the controller and use that in  case of `null`.

DependencyInjection/Configuration.php
dbu: what about adding the defaults here?

ElectricMaxxx: ups

dbu:
@Soullivaneuh do you have a better idea how we can handle the default from the resource bundle?

greg0ire:
> Question @greg0ire How shall i deprecate the configuration on 1.x? Upgrade notes only?

Woops didn't see your question for some reason. It's best to trigger a deprecation error when relevant.

ElectricMaxxx:
@greg0ire but how, and when to trigger a deprecation warnin on a symfony configuration? In  Extension class on 1.x branche?

greg0ire:
Yeah, inside the extension, or maybe inside a `beforeNormalization` node, your call.

DependencyInjection/Configuration.php
```diff
@@ -65,41 +63,20 @@ public function getConfigTreeBuilder()
...
+                        ->scalarNode('repository_name')
+                            ->defaultNull()
+                            ->info('The repository name the resource API connects to.')
                         ->end()
-                        ->scalarNode('precise_children')
-                            ->defaultTrue()
-                            ->info('Exact check if document has children. For large trees, set to false for better performance, but user might needs to click on expand to see there are no children.')
+                        ->arrayNode('routing_defaults')
+                            ->prototype('scalar')->end()
+                            ->info('Routing defaults passed to the resources API call.')
```
greg0ire: Should be moved to #466

ElectricMaxxx: as discussed in the PR, so we can merge this one here?

greg0ire: yes :+1:

ElectricMaxxx: Can you please cause i am the opener of the PR (got some trouble last time *g*)

greg0ire: Can I please what?

ElectricMaxxx: Hit the button?

greg0ire: Sure thing, thanks!

PR #456, opened 2017-07-05
add deprecation notices for configuration changes onto version 2.0

ElectricMaxxx:
notices for #453


ElectricMaxxx:
This is not the change i thought i get. I did a `git mv UPGRADE-1.x.md UPGRADE-2.x.md` to handle the name change, But the diff is something else.

Soullivaneuh:
Why moving from 1.x to 2.x document?

ElectricMaxxx:
@Soullivaneuh cause we (@dbu and me) think that we added the notices how to upgrad from 1.x to 2.x. Is there something different in Sonata?

dbu:
imho we should keep the 1.x document to say the pager is deprecated. the 2.x document then says the pager is removed, repeating the solution from 1.x deprecation.

dbu:
but that PR should go against master. in 1.x all is fine. i would do one pr to move the 2.x infos into their own file in master branch, and then rebase the pull request for configuration changes and add info on changed configuration. (note that they are not deprecated but simply removed. the old config settings are not accepted in 2.x versions)

dbu:
we can't deprecate those configs in 1.x as they are still used. lets just document things in #453 and close this PR, ok?

PR #466, opened 2017-07-27
add deprecation warnings not use some configurations anymore

ElectricMaxxx:
Deprecation warnings for the removal of configuration in #453 
1.2.8 need to be released then
1.x need to merged into master then

### DIFF
--- a/DependencyInjection/SonataDoctrinePHPCRAdminExtension.php
+++ b/DependencyInjection/SonataDoctrinePHPCRAdminExtension.php
@@ -94,10 +94,12 @@ class SonataDoctrinePHPCRAdminExtension extends AbstractSonataAdminExtension
             E_USER_DEPRECATED
         );
 
-        @trigger_error(
-            'The "sonata_doctrine_phpcr_admin.document_tree_defaults" configuration is deprecated since 1.2.8, to be moved in 2.0.',
-            E_USER_DEPRECATED
-        );
+        if (isset($config['document_tree_defaults'])) {
+            @trigger_error(
+                'The "sonata_doctrine_phpcr_admin.document_tree_defaults" configuration is deprecated since 1.2.8, to be moved in 2.0.',
+                E_USER_DEPRECATED
+            );
+        }
 
         $options = $config['document_tree_options'];
         $container->setParameter('sonata_admin_doctrine_phpcr.tree_block.defaults', $config['document_tree_defaults']);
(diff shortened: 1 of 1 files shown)


DependencyInjection/SonataDoctrinePHPCRAdminExtension.php
```diff
@@ -89,6 +89,16 @@ public function getNamespace()
...
     {
+        @trigger_error(
+            'The "sonata_doctrine_phpcr_admin.document_tree_options" configuration is deprecated since 1.2.8, to be replaced in 2.0.',
+            E_USER_DEPRECATED
+        );
+
+        @trigger_error(
+            'The "sonata_doctrine_phpcr_admin.document_tree_defaults" configuration is deprecated since 1.2.8, to be moved in 2.0.',
+            E_USER_DEPRECATED
+        );
```
greg0ire: Shouldn't there be some kind of condition here?

DependencyInjection/SonataDoctrinePHPCRAdminExtension.php
```diff
@@ -89,6 +89,18 @@ public function getNamespace()
...
+            'The "sonata_doctrine_phpcr_admin.document_tree_options" configuration is deprecated since 1.2.8, to be replaced in 2.0.',
+            E_USER_DEPRECATED
+        );
+
+        if (isset($config['document_tree_defaults'])) {
+            @trigger_error(
+                'The "sonata_doctrine_phpcr_admin.document_tree_defaults" configuration is deprecated since 1.2.8, to be moved in 2.0.',
+                E_USER_DEPRECATED
+            );
+        }
```
greg0ire: What should be used instead of the deprecated configuration nodes?

greg0ire: This PR is supposed to introduce the thing you are deprecating this in favor of if it is not already available. The PR on master should only do removals.

DependencyInjection/SonataDoctrinePHPCRAdminExtension.php
```diff
@@ -89,6 +89,20 @@ public function getNamespace()
...
+        @trigger_error(
+            'The "sonata_doctrine_phpcr_admin.document_tree_options" configuration is deprecated since 1.2.8, to be replaced in 2.0.',
+            E_USER_DEPRECATED
+        );
+
+        if (isset($config['document_tree_defaults'])) {
+            @trigger_error(
+                'The "sonata_doctrine_phpcr_admin.document_tree_defaults" configuration is deprecated
+                since 1.2.8, to be moved in 2.0.
+                Use "sonata_doctrine_phpcr_admin.document_tree.routing_defaults" then',
```
greg0ire: You should introduce `sonata_doctrine_phpcr_admin.document_tree.routing_defaults` in this PR and remove " then". Same goes for the other deprecations

dbu:
you can't deprecate something without offering the new solution. as far as i can tell, we can't do forward compatibility from 1.x to 2.0, so there is no point in my opinion to have any deprecations for this. as long as the upgrade instructions are clean, thats good enough for me.

greg0ire:
Thanks for clearing that up @dbu !

PR #468, opened 2017-07-27
File structure

### LINKED ISSUE #457: File structure in tests
I would suggest to create  new structure for the tests and to create Webtests. So lets do something like:

```
Tests/
   Unit/
        ... tests we currently do.
   WebTests/
```

ElectricMaxxx:
fix #457


ElectricMaxxx:
It looks better now.

phpunit.xml.dist
```diff
@@ -9,7 +9,7 @@
          processIsolation="false"
          stopOnFailure="false"
          syntaxCheck="false"
-         bootstrap="Tests/bootstrap.php"
+         bootstrap="tests/bootstrap.php"
```
dbu: @Soullivaneuh does that work? or is this file managed by the dev-kit?

greg0ire: We manage libs that are not bundle and follow this directory structure: https://github.com/sonata-project/GoogleAuthenticator/ . So it should be fine or at least, possible.

ElectricMaxxx: but ...
https://travis-ci.org/sonata-project/SonataDoctrinePhpcrAdminBundle/jobs/258163731

ElectricMaxxx:
@greg0ire you have got no `Resources/docs` folder in your example, that folder is missing now ...

phpunit.xml.dist
dbu: yep. looks like the sphinx build has a hardcoded path. we could keep Resources/doc outside of src/ - while this seems not common practice, it would actually make sense to me, i see no point in putting that inside the source code part.

greg0ire:
I don't know what it should be... @dbu ? `doc` at the top level, or `src/Resources/doc`?

ElectricMaxxx:
:+1:  for `doc` toplevel, but the website should know it too.

greg0ire:
Yeah, and I'm afraid it implies modifying `dev-kit`

ElectricMaxxx:
we could leave `Resources/docs` on top level :-) but that would be ugly

dbu:
> Yeah, and I'm afraid it implies modifying dev-kit
> we could leave Resources/docs on top level :-) but that would be ugly

thats why i suggested Resources/docs :-D
i think we should go with that for now, and see how we move on. until dev-kit would allow a config and the website would look in several places, its quite a hassle to move it elsewhere.

greg0ire:
Go with the ugly solution then, keep the ball rolling :)

ElectricMaxxx:
10 files changed less :-)

greg0ire:
Here are the files managed by dev-kit: https://github.com/sonata-project/dev-kit/tree/master/project

ElectricMaxxx:
docs build is green, this goes completely green now :-)

greg0ire:
green

ElectricMaxxx:
time to merge :-)

greg0ire:
Thanks @ElectricMaxxx !

dbu:
yep, thanks!

PR #470, opened 2017-07-27
Prepare new release

### LINKED ISSUE #449: cleanup composer.json
there are a couple of `@dev` dependencies in composer.json. we have to move to stable versions of the dependencies before releasing.

ElectricMaxxx:
just a clean up before the new release fix #449

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,8 +21,8 @@
         "doctrine/phpcr-odm": "^1.1 || ^2.0",
         "sonata-project/admin-bundle": "^3.1",
         "sonata-project/block-bundle": "^3.0",
-        "symfony-cmf/resource-rest-bundle": "^1.0@dev",
-        "symfony-cmf/tree-browser-bundle": "^2.0@dev",
+        "symfony-cmf/resource-rest-bundle": "^1.0",
+        "symfony-cmf/tree-browser-bundle": "^2.0",
         "symfony/framework-bundle": "^2.8 || ^3.0",
         "symfony/property-access": "^2.8 || ^3.0"
     },


ElectricMaxxx:
argh, seems as i would have to increas coverage before releasing

greg0ire:
No you don't have to, it's not marked as required. You do have to fix the tests though :P

ElectricMaxxx:
Uhh and that is strange. It is just a single build. And there each request on some http tooks to much time. 
Btw why do we download a phpunit.phar, thought we use simple-phpunit?

greg0ire:
[We want to](https://github.com/sonata-project/dev-kit/issues/197), but nobody has made a PR to change that yet.

ElectricMaxxx:
I will give it a try later or tomorrow morning. But the main problem looks more like the container has bad connectivity.

greg0ire:
I don't think so. I think the SAT solver is stuck. Maybe try bumping version constraints to the nearest patch version? @dbu had a similar issue IIRC, solved it using `travis_wait` I think

ElectricMaxxx:
But the test call comes from make file, so that also comes from dev-kit?

greg0ire:
...which helps me find what I was talking about: https://github.com/sonata-project/dev-kit/commit/768f3d66d8b146e9f00c278ef66764b6388bb8bd

ElectricMaxxx:
Oh, i tought to go on RC tonight :-(

greg0ire:
Like I said, You can always try bumping e.g `^1.0` to `^1.0.42`, it could help, because it changes the space of the possibilities composer has to evaluate, but aside from that, sorry I have no solution.

ElectricMaxxx:
So, the solution seems not to work. Or the time is not enough. 

But wait, i released tree-browser-bundle with `min-stability:RC` - unfortunatly. Can that be a reason for composer not to resolve a stable package, which i request here?

greg0ire:
> So, the solution seems not to work. Or the time is not enough.

What makes you say that? The commit from @dbu has not landed on master.

greg0ire:
> But wait, i released tree-browser-bundle with min-stability:RC - unfortunatly. Can that be a reason for composer not to resolve a stable package, which i request here?

No idea, and also, touching that setting or even using @dev in a library does not sound great, even on a master branch. This should only be used in projects IMO, it's a hack.

ElectricMaxxx:
@greg0ire i removed that notice as i saw your answer.

ElectricMaxxx:
The [tree-browser-bundle Release](https://github.com/symfony-cmf/tree-browser-bundle/releases/tag/2.0.1) saved my PR :-)

Ready to merge and RC-1 i would say then.

greg0ire:
Good job!

PR #474, opened 2017-07-31
Fix reorder

ElectricMaxxx:
Server Side Fix for https://github.com/symfony-cmf/tree-browser-bundle/issues/153

### DIFF
--- a/src/Controller/TreeController.php
+++ b/src/Controller/TreeController.php
@@ -102,6 +102,10 @@ class TreeController extends Controller
             return new JsonResponse(array('Parameters parent, dropped and target has to be set to reorder.'), Response::HTTP_BAD_REQUEST);
         }
 
+        if (in_array($position, ['over', 'child'])) {
+            return new JsonResponse(array('Can not resort into a collection.'), Response::HTTP_BAD_REQUEST);
+        }
+
         $before = 'before' == $position;
         $parentNode = $this->session->getNode($parentPath);
         $targetName = PathHelper::getNodeName($targetPath);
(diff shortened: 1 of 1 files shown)


src/Controller/TreeController.php
```diff
@@ -102,6 +102,10 @@ public function reorderAction(Request $request)
             return new JsonResponse(array('Parameters parent, dropped and target has to be set to reorder.'), Response::HTTP_BAD_REQUEST);
         }
 
+        if (in_array($position, array('over', 'child'))) {
+            return new JsonResponse(array('Can not resort into a collection.'), Response::HTTP_BAD_REQUEST);
```
greg0ire: Is "resort" really what you mean here? http://www.wordreference.com/enfr/resort
the name of the action is "reorder" BTW

ElectricMaxxx: i mean "reorder", and that should be done, when position is `before` or `after` a node. Reorder will interally set the position property of a node collection.

ElectricMaxxx: now i see your issue, i am not that good in wording, even in my home language :-) So what do i mean, based on the description before?

greg0ire: Then use reorder in the message please

ElectricMaxxx: thanks. BTW: this route will be removed, when `resources-rest-bundle` provides a PATCH route  on our position.

ElectricMaxxx:
i will load that code into our sandbox and see it working before merging and releasing.

ElectricMaxxx:
i excluded that call on client side, but whanted to have the backend fail safe.

PR #477, opened 2017-08-04
Don't merge, proof PR : Fix travis wait

greg0ire:


greg0ire:
The issue with this solution is that it makes logs very shitty. Should be used very sparingly.

greg0ire:
better solution: #479

PR #479, opened 2017-08-05
Don't merge, POC PR: Fix travis wait again.

jlamur:


greg0ire:
Will not be needed after all, I discovered why Composer was taking so long :)

PR #482, opened 2017-08-05
Stop testing Sonata 2.x

greg0ire:
Sonata 2 is no longer maintained.


ElectricMaxxx:
We should mention, that we also stop to support Sonata 2.x

greg0ire:
Let's do that in another PR, ok?

PR #489, opened 2017-08-23
Fix tree root node

EmmanuelVella:
## Changelog

```markdown
### Fixed
- Missing root_node parameter in cmfTree options
```

## Subject

Hi ! I try to update to cmf 2.0. I'm not comfortable with the tree JS, but I think there is an issue with the cmfTree root_node parameter. Here are the results before/after the fix :

| Before  | After |
| ------------- | ------------- |
| before | after |

### DIFF
--- a/src/Resources/views/Tree/tree.html.twig
+++ b/src/Resources/views/Tree/tree.html.twig
@@ -24,11 +24,11 @@ file that was distributed with this source code.
 <script type="text/javascript">
     var apiGetTemplate = '{{ path('_cmf_get_resource', {
         repositoryName: repository_name,
-        root_node: root_node,
         path: '__path__'
     }|merge(routing_default_values)) }}';
     jQuery(function($) {
         $('#tree').cmfTree({
+            root_node: '{{ root_node }}',
             request: {
                 load: function (nodePath) {
                     return {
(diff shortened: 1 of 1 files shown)


EmmanuelVella:
ping @ElectricMaxxx

src/Resources/views/Tree/tree.html.twig
```diff
@@ -24,11 +24,11 @@ file that was distributed with this source code.
 <script type="text/javascript">
     var apiGetTemplate = '{{ path('_cmf_get_resource', {
         repositoryName: repository_name,
-        root_node: root_node,
```
ElectricMaxxx: I think we still need that node here, isn't it?

EmmanuelVella: I don't think so, the route and the controller do not use it : 
https://github.com/symfony-cmf/resource-rest-bundle/blob/039884f80d2864f6457da5988801ae42cad439a9/src/Resources/config/routing.yml#L17
https://github.com/symfony-cmf/resource-rest-bundle/blob/master/src/Controller/ResourceController.php#L68

ElectricMaxxx: now it makes sense, sorry i was sleeping. Can you remove it in Line 43, too. best would be to use `apiGetTemplate.replace('__path__', sourcePath),` there.

EmmanuelVella: Done !

dbu:
thank you!

PR #491, opened 2017-08-29
Remove exceptions from phpunit whitelist

jordisala1991:
I am targeting this branch, because this is pedantic.

## Subject

I think coverage is not calculated correctly, lets see

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -12,21 +12,26 @@
          bootstrap="Tests/bootstrap.php"
 >
     <testsuites>
-        <testsuite name="PHPCR Util Test Suite">
-            <directory>./Tests/</directory>
+        <testsuite name="SonataDoctrinePHPCRAdminBundle Test Suite">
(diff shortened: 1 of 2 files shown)


phpunit.xml.dist
```diff
@@ -12,21 +12,26 @@
          bootstrap="Tests/bootstrap.php"
 >
     <testsuites>
-        <testsuite name="PHPCR Util Test Suite">
-            <directory>./Tests/</directory>
+        <testsuite name="DashboardBundle Test Suite">
```
OskarStark: wrong name

OskarStark:
@ElectricMaxxx how do we proceed with version eye? can we whitelist the license?

dbu:
regarding version eye i have no idea how to tell it that we are ok with depending on PHP in a PHP project :-( i clicked around in the GUI but did not even find sonata in the projects list.

OskarStark:
Thank you for checking this @dbu 👍

PR #499, opened 2017-10-14
Drop old versions of php/sf

core23:
I am targeting this branch, because this is BC.

## Changelog

```markdown
### Removed
- Support for old versions of PHP and Symfony.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,15 +16,15 @@
         }
     ],
     "require": {
-        "php": "^5.6 || ^7.0",
+        "php": "^7.1",
         "doctrine/phpcr-bundle": "^1.1",
         "doctrine/phpcr-odm": "^1.3 || ^2.0",
         "sonata-project/admin-bundle": "^3.21.0",
         "sonata-project/block-bundle": "^3.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -36,8 +36,8 @@
         "symfony-cmf/resource": "^1.0",
         "symfony-cmf/resource-bundle": "^1.0",
         "symfony-cmf/testing": "^2.0",
-        "symfony/phpunit-bridge": "^2.8 || ^3.0",
-        "symfony/security-acl": "^2.8 || ^3.0"
+        "symfony/phpunit-bridge": "^3.3",
+        "symfony/security-acl": "^2.8 || ^3.2"
```
jordisala1991: This should be ^3.0

core23: Fixed

jordisala1991: it is still 3.0

core23: Looks like a github Problem oO
https://github.com/core23/SonataDoctrinePhpcrAdminBundle/blob/patch/php-support/composer.json#L40

jordisala1991:
It does not let me reopen it :S

PR #506, opened 2017-10-25
DevKit updates for 2.x branch

SonataCI:
## Changelog
```markdown
### Changed
- Rollback to PHP 5.6 as minimum support.
- getSubject() implementation calls `AbstractAdmin::getObject()` now to avoid calling `getClass()`
```


dbu:
seems to need some travis_wait to not time out with php 5.6

greg0ire:
@dbu maybe there are some branches that could be deleted, I'm under the impression you guys have been not using forks again.

greg0ire:
For instance in phpcr_bundle:

dev-allow_symfony_4.0
dev-test-php-72
dev-cache_queries

in ODM:


dev-test-php-7
dev-cascade_children
dev-find_translation_reg
dev-htree-ops

in tree browser bundle:


dev-fix_installation
dev-autocomplete

etc.

greg0ire:
cc @ElectricMaxxx can you help with the branches?

ElectricMaxxx:
Mhh. Same we had a half a year ago.

.travis.yml
```diff
@@ -45,25 +47,23 @@ matrix:
     - php: '7.1'
       env: SYMFONY=3.3.*
     - php: '7.1'
-      env: SYMFONY=dev-master@dev
+      env: SYMFONY='dev-master as 3.3'
```
dbu: dev-master of symfony is 4.1 at this point, i think. we should however test with 3.4 and 4.0 eventually... but lets not do too many things in the same PR.

composer.json
```diff
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
-        "php": "^7.1",
+        "php": "^5.6 || ^7.0",
```
dbu: this looks like a regression to me

greg0ire: it is not

dbu:
regarding branches: argh, we seem to forget that again... its a different workflow from what i was used to. did it break installation?

greg0ire:
I'm really not sure this time

composer.json
dbu: ok, learned that its on purpose. but there is no point in allowing those as new phpcr-odm versions only support php 7.1 as well.

greg0ire: If that's true, the 5.6 build should fail

greg0ire: So did you drop versions of php-odm that allow 5.6 or not?

dbu: hm, i guess the sonata bundle still supports phpcr-odm 1.x so it does make sense to keep this in, at least for now.

.travis.yml
ElectricMaxxx: We should test with 3.4 directly instead of aliasing. Doing so we see the issues we have to fix and issues can also be a depending package that still does not like 3.4/4.0

ElectricMaxxx:
I tried it as [Nils|https://twitter.com/naderman/status/933308769706106880] said and got:
```
Resolving dependencies through SAT
Dependency resolution completed in 185.155 seconds
Analyzed 12560 packages to resolve dependencies
Analyzed 1583412 rules to resolve dependencies
Resolving dependencies through SAT
Dependency resolution completed in 0.000 seconds
Package operations: 3 installs, 56 updates, 9 removals
```
That does not look that bad. ...

greg0ire:
Well I'm restarting the job in case you guys did some cleanup and if fixed it, let's see...

jordisala1991:
What PHP version are you using @ElectricMaxxx to profile composer? The issue is with 5.6, on 7.0-7.1 there is no problem

greg0ire:
This repo is full of dev branches: https://packagist.org/packages/symfony-cmf/resource-bundle

ElectricMaxxx:
@jordisala1991 ah, yes you are right :-) i have no 5.6 anymore
wait i got the box of my customer with 5.6 on a ugly centos. i will try it there tomorrow

ElectricMaxxx:
I removed 4 of 9 branches from `resouce-bundle`lets see what happens.

jordisala1991:
Still doesnt pass. What shows you composer profile on 5.6?

greg0ire:
This job errors, and now the other builds are failing :(

greg0ire:
Ok let me try to fix this.

greg0ire:
Unfixable w/o https://github.com/sonata-project/SonataAdminBundle/pull/4824

jordisala1991:
I did a squash on the dev-kit commits, and added some necessary things that will be needed once the problem with the endless loop gets fixed

jordisala1991:
Now we are getting another error?

```
1) Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin\ContentAdminTest::testContentList

Error: Call to undefined method Symfony\Component\Yaml\Parser::parseFile()
```

greg0ire:
I believe this is an autoloading conflict between our dependencies and those of phpunit

greg0ire:
I reproduce the issue locally, and `parseFile` does exist.

greg0ire:
Any way, not looking into this until there's a new release of the admin bundle

dbu:
that yaml thing sounds like one of the reasons the symfony phpunit bridge exists. you can specify to the bridge that it should remove some of the phpunit dependencies so the ones from the project are used. https://symfony.com/doc/3.4/components/phpunit_bridge.html might help?

greg0ire:
@dbu yes it could, and I considered it at some point. Maybe it's time for me to get back to this: https://github.com/sonata-project/dev-kit/pull/265

jordisala1991:
We are back to the start point. Build is failing cause --prefer-lowest takes more than 10min to install composer dependencies

greg0ire:
Maybe add the `travis_wait` directly on this branch for now, and deal with that later (or never, upgrade to 7.1 might be not that far)

jordisala1991:
Done... let's try with travis_wait...

IMO having to install symfony-cms/testing which requires directly `symfony/symfony` instead of individuals components/bundles/bridges hurts a lot on composer resolution.

greg0ire:
Dichotomy might be useful in this situation: remove half the deps, add a quarter of them back, rinse and repeat until you pinpoint the dep that is causing the issue

jordisala1991:
Aaaaand we are done here :) finally

greg0ire:
Great job @jordisala1991 !!!

dbu:
yay, thanks a lot!

PR #509, opened 2017-11-16
Compatibility

ElectricMaxxx:
I am targeting this branch, because {reason}.

Cause i want to use it in next version.
## Changelog
```Markdown
### Added
- version 2.0 for `phpcr-odm`
```
## Subject

I will add step by step changes to get symfony 4.0 running on cmf bundles and so here too, but i will do it step by step in separate PRs.

PR #514, opened 2018-01-03
DevKit updates for master branch

SonataCI:


jordisala1991:
Also, @dbu can you check if we can remove a branch called `compatibility`?

dbu:
@ElectricMaxxx created the commits in the `compatibility` branch. max, are those commits still relevant or can we remove the branch?

dbu:
there is #509 that was merged, but there the compatibility branch is in @ElectricMaxxx fork of this repo... i guess we can delete the branch then, it probably got accidentally pushed to upstream?

jordisala1991:
@dbu do you know why this happens? Can't we have an entity with `:void` as a return type on a method? I mean, I can force them all to return this so we don't get bother by this, but doesn't seem a solution to me.

https://travis-ci.org/sonata-project/SonataDoctrinePhpcrAdminBundle/jobs/326118558#L845

dbu:
in OO, an extending class my only narrow the return type of a method to a subclass of what the parent declares. PHP does not implement that afaik. but afaik no language would allow to widen the type. the proxies use inheritance, hence they must also do the return types. as this only happens on the lowest build, it must be some dependency that is installed in a too old version - my guess would be the proxy generator.

jordisala1991:
Proxy generator is a dependency directly declared here or it is from another bundle?
If it is on another bundle I would need to update that dependency and release that other bundle to make this work?

dbu:
its no direct dependency. i tried finding what version is installed in a green vs the red build, but can't find that in the build log. i guess we should bump the dependency on phpcr-odm to 2.0

jordisala1991:
I think the package that causes this issue is doctrine-common

dbu:
> I think the package that causes this issue is doctrine-common

probably. can you find which version adds the return type declaration support and we add a conflict with versions lower than that?

jordisala1991:
https://github.com/symfony-cmf/testing/pull/169

This one will fix this build

jordisala1991:
Aaaand it seems to be working finally :)

jordisala1991:
```
[PHPCR] Error when initializing dbal: 
In CheckExceptionOnInvalidReferenceBehaviorPass.php line 32:
                                                                               
  The service "jms_serializer.stopwatch_subscriber" has a dependency on a non  
  -existent service "debug.stopwatch".                                         
```

And now another error, this time is different, and it has no sense because all they builds passing but the admin 3.x is not

dbu:
do we install a different version of jms serializer or (i guess) symfony/debug in the build that fails? or a different framework bundle? those would be the things that are involved with those services. or a different version of jackalope or doctrine dbal that does something different with the serializer?

jordisala1991:
The only difference is that we use the 3.x branch instead of a tag, but that shouldnt change things... or atleast I dont remember any change affected by this...

dbu:
do you manage to reproduce the problem locally? that might give some indication of what is happening...

ElectricMaxxx:
@dbu is that question still relevant for me? I can not find any branch with that name atm?

jordisala1991:
I removed it because of what dbu said

dbu:
i merged 2.x to master and now rebased this branch on master. lets see!

jordisala1991:
Done, thank you @dbu

PR #524, opened 2018-05-06
DevKit updates for 2.x branch

SonataCI:


dbu:
the build is red because of the --prefer-lowest build. i guess thats not caused by the changes in this pull request. should we merge it?

PR #521, opened 2018-01-23
support Symfony 4

ElectricMaxxx:
I am targeting this branch, because i want to work on a new 2.x version

## Changelog

```markdown
### Added
- Symfony 4 support
```

## Subject

I want to allow Symfony 4.0 before making it required in travis.yaml, just to installi it in a testing project and to see what happens. I would add outcome as issues.

PR #528, opened 2018-08-28
commit all cs issues agianst our master here 2.x

ElectricMaxxx:
I am targeting this branch, because In #521 we do need a clean "master" (mean 2.x).


ElectricMaxxx:
We should not be that carefully with the red tests as i fixed them in #521

dbu:
thanks!

PR #512, opened 2017-12-05
Avoid loop

ElectricMaxxx:
I am targeting this branch, because i face the loop in here

## Changelog

```markdown
### Changed
- getSubject() implementation calls `AbstractAdmin::getObject()` now to avoid calling `getClass()`
```

## Subject

In our test we hit a loop, cause we had to implement getSubject() on our own, where we used getClass(), which calls getSubject() again.

PR #535, opened 2018-11-12
PHP 7.1 cleanups

dbu:
Some leftover from #521 

## Subject

Cleanup tests a bit with :void return type declarations.


docs/reference/document_tree.rst
```diff
@@ -220,7 +220,7 @@ like this:
     .. code-block:: jinja
 
         {% render(controller(
-            'sonata.admin.doctrine_phpcr.tree_controller:treeAction',
+            'sonata.admin.doctrine_phpcr.tree_controller::tree',
```
dbu: don't we still need the "Action" suffix with the double colon notation?

dbu: according to https://symfony.com/doc/current/best_practices/controllers.html#controller-action-naming the `Action` suffix is deprecated. however, the thing here must match the method name, and you are not adjusting the method name in this PR.

changing the method name is a BC break. i suggest we make a ticket here for this, first adding the method and make the `Action` methods call those method and trigger a deprecation warning, then later remove it when we do the next major version. but its a bit out of scope of this PR, and this PR is big enough as is.

OskarStark:
@dbu there are a lots of commits, can you please get rid of them?

Or apply these changes against the PR of @ElectricMaxxx so his PR gets updated, which could then be merged, and we don't los the comments there.

dbu:
yeah, was just thinking the same :-D i will squash that thing

ElectricMaxxx:
@dbu can ammend a more meaningfull commit message pls?

ElectricMaxxx:
Ah, the last commit wasn't the squash result :-)

but at the end we should have one commit  explaining the whole change :-)

dbu:
as #512, this PR goes against the 2.x branch, but it contains commits that are only in master. i guess i throw out those master commits in the rebase?

.gitattributes
```diff
@@ -2,3 +2,4 @@
 *.md export-ignore
 tests/* export-ignore
 docs/* export-ignore
+tests/Fixtures/App/var* export-ignore
```
OskarStark: This one gets overwritten by `dev-kit` on the next run:
https://github.com/sonata-project/dev-kit/blob/master/project/.gitattributes.twig

.gitignore
```diff
@@ -2,7 +2,8 @@
 /docs/_build
 /vendor
 .php_cs.cache
+.phpunit
```
OskarStark: do we need this?

.travis/check_relevant_lint.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
```
OskarStark: This one gets overwritten to `bash` by `dev-kit` on the next run.

Why did you change this?

.travis/check_relevant_docs.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
```
OskarStark: This one gets overwritten to `bash` by `dev-kit` on the next run.

Why did you change this?

.travis/check_relevant_test.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
```
OskarStark: This one gets overwritten to `bash` by `dev-kit` on the next run.

Why did you change this?

.travis/before_script_test.sh
```diff
@@ -1,3 +1,10 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
```
OskarStark: This one gets overwritten to `bash` by `dev-kit` on the next run.

Why did you change this?

.travis/before_script_test.sh
```diff
@@ -1,3 +1,10 @@
-#!/usr/bin/env bash
```
OskarStark: Please revert the CHMOD

.travis/before_script_test.sh
```diff
@@ -1,3 +1,10 @@
...
+#!/usr/bin/env sh
 set -ev
 
+TESTING_SCRIPTS_DIR=vendor/symfony-cmf/testing/bin
+CONSOLE=${TESTING_SCRIPTS_DIR}/console
+
+export KERNEL_CLASS="Sonata\\DoctrinePHPCRAdminBundle\\Tests\\Fixtures\\App\\Kernel"
+echo '+++ create PHPCR +++'
+${CONSOLE} doctrine:phpcr:init:dbal --drop --force -vvv
+${CONSOLE} doctrine:phpcr:repository:init -vvv
```
OskarStark: The whole block gets overwritten by `dev-kit` on the next run.

This should be added [here](https://github.com/sonata-project/dev-kit/blob/master/project/.travis/before_script_test.sh.twig) with a condition on the project.

phpunit.xml.dist
```diff
@@ -11,19 +11,11 @@
          syntaxCheck="false"
```
OskarStark: as `phpunit.xml.dist` file is managed by `dev-kit`, I am not sure what will survive after the next run

docs/reference/document_tree.rst
dbu: created #537 and changed the doc back to use `Action`

dbu:
there we go. squashed the thing and removed the 3.x commit. please look over it again and merge if you are happy.

dbu:
@ElectricMaxxx would you mind porting the CI changes to the dev-kit?

ElectricMaxxx:
@dbu that is quite a good question, as i heard it isn't possible to have separate travis scripts in sonata-dev-kit, isn't it?

OskarStark:
You can do this, we are doing it here for example:
https://github.com/sonata-project/dev-kit/blob/master/project/.travis/before_install_test.sh.twig

.travis/before_script_test.sh
OskarStark: You are right, I thought it was done by mistake.

We can keep this, I added a PR on dev-kit for the other bundle 👍 
https://github.com/sonata-project/dev-kit/pull/440

dbu:
@ElectricMaxxx i cleaned this up again. can you maybe port the things we need to be in the dev-kit there so that we can merge this and won't lose changes on the next dev-kit run?

.travis/before_script_test.sh
ElectricMaxxx: thougt dev-kit still got it in?

ElectricMaxxx:
I created: https://github.com/sonata-project/dev-kit/pull/458

.gitattributes
OskarStark: please add this to the dev-kit PR, too

phpunit.xml.dist
ElectricMaxxx: no it isn't :-)

ElectricMaxxx: https://github.com/sonata-project/dev-kit/blob/master/config/projects.yml#L189

OskarStark: You are right 👍

dbu:
i rebased now that #521 is merged and this is all that is left. mostly re-adding the `: void` to test methods. not sure if its worth it but its correct, and for tests there are no BC concerns.

OskarStark:
Can you please update the PR header and the title accordingly?

dbu:
done. i removed the changelog entry, i think this is not worth putting anything into the changelog ;-)

tests/Fixtures/App/DataFixtures/Phpcr/LoadTreeData.php
```diff
@@ -20,7 +20,7 @@
 
 class LoadTreeData implements FixtureInterface
 {
-    public function load(ObjectManager $manager)
+    public function load(ObjectManager $manager): void
```
greg0ire: Isn't this a BC-break for extending classes?

OskarStark: its a fixture

greg0ire:
Thanks @dbu !

PR #540, opened 2018-11-29
DevKit updates for 2.x branch

SonataCI:


greg0ire:
Finally! @core23 I had to switch to `weak` instead of `weak_vendors` to have this pass, because `weak_vendors` wrongly thinks that deprecations coming from `test/app/cache` are non-vendor deprecations, when in fact they are deprecations about core bundle services. dev-kit will re-add the mode, so it would be great to use new service, but I don't know which should be used (the deprecation message does not say in favor of what each service is deprecated). Can you please help with this?

PR #544, opened 2019-01-16
DevKit updates for 2.x branch

SonataCI:


OskarStark:
@greg0ire, currently we use `weak` here, but this PR introduces `weak_vendor`. Which should we use?

Afterwards the change needs to be applied on dev-kit

greg0ire:
We should use `weak_vendors` although it mistakenly treats deprecations coming from a cache in `tests/app/cache` or something as internal deprecations. We should fix the 3 first deprecations, but they say this:
```
  6x: The "sonata.core.model.adapter.chain" service is deprecated since 3.x and will be removed in 4.0.
    1x in ContentAdminTest::testContentList from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in ContentAdminTest::testContentWithChildEdit from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in ContentAdminTest::testContentWithChildrenEdit from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in ContentAdminTest::testContentCreate from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in ContentAdminTest::testShowContent from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in TreeBrowserTest::testTreeOnDashboardLoadsWithNoErrors from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Dashboard
  6x: The "sonata.core.model.adapter.doctrine_orm" service is deprecated since 3.x and will be removed in 4.0.
    1x in ContentAdminTest::testContentList from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in ContentAdminTest::testContentWithChildEdit from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in ContentAdminTest::testContentWithChildrenEdit from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in ContentAdminTest::testContentCreate from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in ContentAdminTest::testShowContent from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in TreeBrowserTest::testTreeOnDashboardLoadsWithNoErrors from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Dashboard
  6x: The "sonata.core.model.adapter.doctrine_phpcr" service is deprecated since 3.x and will be removed in 4.0.
    1x in ContentAdminTest::testContentList from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in ContentAdminTest::testContentWithChildEdit from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in ContentAdminTest::testContentWithChildrenEdit from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in ContentAdminTest::testContentCreate from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in ContentAdminTest::testShowContent from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Admin
    1x in TreeBrowserTest::testTreeOnDashboardLoadsWithNoErrors from Sonata\DoctrinePHPCRAdminBundle\Tests\WebTest\Dashboard
```

@core23 , please tell us in favor of what these 3 services are deprecated

greg0ire:
We could revert that change in this PR until there is a resolution.

OskarStark:
Please merge if green, otherwise dev kit will revert my change in the night 😊

.php_cs.dist
```diff
@@ -43,10 +43,9 @@ $rules = [
     '@PHP71Migration:risky' => true,
     'compact_nullable_typehint' => true,
     'void_return' => null,
-    // To be tested before insertion:
-//    'strict_comparison' => true,
-//    'strict_param' => true,
-//    'php_unit_strict' => true,
+    'strict_comparison' => true,
```
dbu: this is creating a bunch of problems in the tests. some might indicate sloppy code, but others might be plane wrong (assertSame vs assertEquals). can we kick this option again?

OskarStark: No we should fix the tests, we did this across all other bundles

dbu: agree for the numeric things, but where we have objects that have the same value but are not the identical object, why would we not use assertEquals over some hack to avoid doing assertSame?

OskarStark: If it is not the identical object we cannot use assertSame obviously, but in this case asserEquals is wrong too, in this case we should assert, what we want to really assert, for example a return value of a getter

WDYT?

cc @greg0ire

greg0ire: I agree with you, using assertEquals with object feels hackish to me, just like I wouldn't use `==` to compare two dates (or anything, really). For dates, we could check that the difference between them is zero seconds like adviced here: https://github.com/sebastianbergmann/phpunit/issues/2456#issuecomment-277692041

.editorconfig
```diff
@@ -7,7 +7,7 @@ charset = utf-8
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[*.{yml,twig,php}]
+[*.{yaml,yml,twig,php}]
```
ElectricMaxxx: why a second?

OskarStark: Otherwise the editor would not use the options for the new symfony config flies which ends with .yaml instead of .yml

Or what do you mean?

.php_cs.dist
OskarStark: Currently I don’t have that much time, but I fixed all the other bundles (with @greg0ire) and would fix the tests here, too when I find some time...

PR #558, opened 2019-09-30
Fix Child "_delete" does not exist.

sad270:
## Subject
If current role is not allowed to delete the item the `_delete` checkboxes are not added in items.
And when we submit the form we have `Child "_delete" does not exist.` error.

I am targeting this branch, because this is a patch with BC.

## Changelog

```markdown
### Fixed
- `Child "_delete" does not exist` in Collection when we don't have delete permission
```


phansys:
Related to sonata-project/SonataAdminBundle#5532.

greg0ire:
Thanks @sad270 !

PR #562, opened 2019-11-11
Dropping support for symfony < 3.4

franmomu:
## Subject

This is the same thing as https://github.com/sonata-project/SonataAdminBundle/pull/5733, dropping support of Symfony < 3.4 and >= 4, < 4.2

I am targeting this branch, because these changes respect BC.

## Changelog

```markdown
### Removed
- Support for Symfony < 3.4
- Support for Symfony >= 4, < 4.2
```


OskarStark:
3.4 is LTS and still supported

dbu:
okay, that stays supported for quite a while indeed. does not mean new versions of this bundle must support it, but if there is no big benefit then we should not drop support for no reason, agreed.

src/Builder/FormContractor.php
```diff
@@ -113,16 +119,13 @@ public function getDefaultOptions($type, FieldDescriptionInterface $fieldDescrip
...
         switch ($type) {
-            case 'Sonata\DoctrinePHPCRAdminBundle\Form\Type\TreeModelType':
-            case 'doctrine_phpcr_odm_tree':
+            case TreeModelType::class:
                 $options['class'] = $fieldDescription->getTargetEntity();
                 $options['model_manager'] = $fieldDescription->getAdmin()->getModelManager();
 
                 break;
-            case 'Sonata\AdminBundle\Form\Type\ModelType':
-            case 'sonata_type_model':
```
core23: Not sure if we should handle this as a BC break. Using form aliases is deprecated since symfony <= 3, but we are about to remove functionality.

OskarStark:
> okay, that stays supported for quite a while indeed. does not mean new versions of this bundle must support it, but if there is no big benefit then we should not drop support for no reason, agreed.

Lets drop 3.4 in the next major of our bundle, but I am against dropping an already supported one in a minor

composer.json
```diff
@@ -55,11 +56,11 @@
         "sonata-project/core-bundle": "^3.14",
         "symfony-cmf/resource": "^1.0",
         "symfony-cmf/resource-bundle": "^1.0",
-        "symfony-cmf/testing": "^2.1.12",
+        "symfony-cmf/testing": "dev-master as 2.1.12",
```
dbu: should we tag version 3 of testing? it seems the missing piece is https://github.com/symfony-cmf/Testing/issues/166 but we could maybe ignore that for now.

franmomu: I'll better update to `"symfony-cmf/testing": "^3@dev"` and I'd leave it like that for now

greg0ire: @dbu please do :pray: an extra release never hurts, relying on a dev version can.

dbu:
i merged the cleanup of dependencies to the testing component and have restarted the lowest and latest build. if they are green, i will tag 3.0 of testing and then we can reference a stable testing component.

dbu:
now we see that this bundle depends on DoctrineBundle but did not correctly declare that dependency... @franmomu are you up to try to get the dependencies correct, now that symfony-cmf/testing no longer magically adds them? https://github.com/symfony-cmf/Testing/pull/199 might be helpful to see what changed.

franmomu:
> now we see that this bundle depends on DoctrineBundle but did not correctly declare that dependency... @franmomu are you up to try to get the dependencies correct, now that symfony-cmf/testing no longer magically adds them? [symfony-cmf/Testing#199](https://github.com/symfony-cmf/Testing/pull/199) might be helpful to see what changed.

I'll give it a try

composer.json
```diff
@@ -55,11 +58,14 @@
         "sonata-project/core-bundle": "^3.14",
         "symfony-cmf/resource": "^1.0",
         "symfony-cmf/resource-bundle": "^1.0",
-        "symfony-cmf/testing": "^2.1.12",
+        "symfony-cmf/testing": "^3@dev",
```
dbu: i just tagged 3.0.0 now. can you please change this constraint to `^3.0` without the `@dev`?

PR #585, opened 2020-03-24
Make filters behaviour more consistent with others ORM

VincentLanglet:
## Subject

I try to add consistency between SonataDoctrine bundle.

I am targeting this branch, because this should be only
- bugfix
- new feature
- remove deprecation

## Changelog

Will be updated.

```markdown
### Added
BooleanFilter 'operator_type' and 'operator_options' are now overridable

### Fixed
Fixed ChoiceFilter query
```

Todo:
- Update the NodeNameFilter: I don't know is this and the behaviour is weird. Here, https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/blob/2.x/src/Filter/NodeNameFilter.php#L19, it does only implement 2 constants of the ChoiceType. But 4 are available. This would mean that using this filter with `NOT_CONTAINS` has the same effect that using it with `CONTAINS`.
- Update the StringFilter: https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/blob/2.x/src/Filter/StringFilter.php

Both of these Filter use a custom ChoiceType: 
https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/blob/2.x/src/Form/Type/Filter/ChoiceType.php
- This ChoiceType introduce a new constant 
- It's using the translator https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/blob/2.x/src/Form/Type/Filter/ChoiceType.php#L46 which is deprecated in the original ChoiceType and will be removed in 4.0 https://github.com/sonata-project/SonataAdminBundle/blob/3.x/src/Form/Type/Filter/ChoiceType.php#L49

So what to do with this custom ChoiceType ? Dropping it in the next major ? Improving the SonataAdmin one and use it instead ?

### DIFF
--- a/src/Filter/BooleanFilter.php
+++ b/src/Filter/BooleanFilter.php
@@ -14,8 +14,10 @@ declare(strict_types=1);
 namespace Sonata\DoctrinePHPCRAdminBundle\Filter;
 
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
+use Sonata\AdminBundle\Form\Type\Filter\DefaultType;
 use Sonata\DoctrinePHPCRAdminBundle\Filter\Filter as BaseFilter;
 use Sonata\Form\Type\BooleanType;
+use Symfony\Component\Form\Extension\Core\Type\HiddenType;
 
 class BooleanFilter extends BaseFilter
 {
@@ -44,7 +46,11 @@ class BooleanFilter extends BaseFilter
      */
     public function getDefaultOptions()
     {
-        return [];
+        return [
+            'field_type' => BooleanType::class,
+            'operator_type' => HiddenType::class,
+            'operator_options' => [],
+        ];
     }
 
     /**
@@ -52,11 +58,11 @@ class BooleanFilter extends BaseFilter
      */
     public function getRenderSettings()
     {
-        return ['sonata_type_filter_default', [
+        return [DefaultType::class, [
             'field_type' => $this->getFieldType(),
             'field_options' => $this->getFieldOptions(),
-            'operator_type' => 'hidden',
-            'operator_options' => [],
+            'operator_type' => $this->getOption('operator_type'),
+            'operator_options' => $this->getOption('operator_options'),
             'label' => $this->getLabel(),
         ]];
     }
--- a/src/Filter/ChoiceFilter.php
+++ b/src/Filter/ChoiceFilter.php
@@ -14,7 +14,8 @@ declare(strict_types=1);
 namespace Sonata\DoctrinePHPCRAdminBundle\Filter;
 
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
-use Sonata\AdminBundle\Form\Type\Filter\ChoiceType;
+use Sonata\AdminBundle\Form\Type\Filter\DefaultType;
+use Sonata\AdminBundle\Form\Type\Operator\EqualOperatorType;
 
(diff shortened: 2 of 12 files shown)


src/Filter/ChoiceFilter.php
```diff
@@ -45,15 +46,17 @@ public function filter(ProxyQueryInterface $proxyQuery, $alias, $field, $data)
             return;
         }
 
-        $andX = $this->getWhere($proxyQuery)->andX();
```
VincentLanglet: I feel like this was doing

```
 foo LIKE value1 AND foo LIKE value2 AND ...
```

Instead 
```
 foo = value1 OR foo = value2 OR ...
```

src/Filter/ChoiceFilter.php
```diff
@@ -45,15 +46,17 @@ public function filter(ProxyQueryInterface $proxyQuery, $alias, $field, $data)
...
+        if (EqualOperatorType::TYPE_NOT_EQUAL === $type) {
+            $where = $this->getWhere($proxyQuery)->andX();
+        } else {
+            $where = $this->getWhere($proxyQuery)->orX();
+        }
 
         foreach ($values as $value) {
-            if (ChoiceType::TYPE_NOT_CONTAINS === $type) {
```
VincentLanglet: The `if` was using the 3 ChoiceType constants, but the `operator_type` was the EqualType, which provide just 2 constants.

src/Filter/ChoiceFilter.php
```diff
@@ -66,16 +69,20 @@ public function filter(ProxyQueryInterface $proxyQuery, $alias, $field, $data)
...
 
     /**
      * {@inheritdoc}
      */
     public function getRenderSettings()
     {
-        return ['sonata_type_filter_default', [
-            'operator_type' => 'sonata_type_equal',
+        return [DefaultType::class, [
+            'operator_type' => $this->getOption('operator_type'),
```
VincentLanglet: This is a new feature in DoctrineORM, the ability to override the operator_type (to use an hiddenType for example).

src/Filter/BooleanFilter.php
```diff
@@ -44,19 +46,23 @@ public function filter(ProxyQueryInterface $proxyQuery, $alias, $field, $data)
...
      */
     public function getRenderSettings()
     {
-        return ['sonata_type_filter_default', [
+        return [DefaultType::class, [
             'field_type' => $this->getFieldType(),
             'field_options' => $this->getFieldOptions(),
-            'operator_type' => 'hidden',
-            'operator_options' => [],
+            'operator_type' => $this->getOption('operator_type'),
```
VincentLanglet: This is a new feature in DoctrineORM

VincentLanglet:
@sonata-project/contributors I'll be happy to have some opinion/advice about this started PR :)

VincentLanglet:
@core23 @phansys @greg0ire I'll be happy to have some opinion on the NodeNameFilter and the StringFilter.

I don't understand what is doing the NodeNameFilter. 
What is the point in having 4 operators if 3 of them are doing the same query.

The StringFilter has a supplementary option for the operator, which is actually not handle by other SonataORMBundle.

Both of these filter use a custom ChoiceType, which extends the Sonata\AdminBundle\Form\Type\Filter\ChoiceType but the SonataChoiceType will be final in 4.0.

Should we drop this custom ChoiceType for next major ?

PR #590, opened 2020-04-14
Applied fixes from FlintCI

soullivaneuh:


src/Datagrid/ProxyQuery.php
```diff
@@ -60,8 +60,6 @@ class ProxyQuery implements ProxyQueryInterface
 
     /**
      * PHPCR ODM Document Manager.
-     *
-     * @var DocumentManager;
```
dbu, line 64: should this really be removed?

OskarStark: Maybe it’s spotted because it ends with a semicolon?

dbu: ok the semicolon sounds wrong. might be https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4913 - though here i think the fix would be to remove the semicolon

OskarStark: I fixed it in `2.x`, lets close this one, so we get a new fresh one from afterwards...

PR #601, opened 2020-07-22
Mark the project as abandoned

VincentLanglet:


jordisala1991:
I think we need to remove it from dev-kit too, otherwise it will update again the readme removing this

dbu:
> I guess we could also add "abandoned": true in composer.json.

does that set the abandoned option on packagist? otherwise i can do that manually once this is merged

core23:
> does that set the abandoned option on packagist? otherwise i can do that manually once this is merged

Setting it inside the `composer.json` will abandoned in on packagist.

phansys:
>does that set the abandoned option on packagist? otherwise i can do that manually once this is merged

I guess that will be the result, but I can't confirm since there is no reference to Packagist at [docs](https://getcomposer.org/doc/04-schema.md#abandoned).
Regardless that, I think is better to have the `abandoned` definition set directly in the package. IMO, that way the package is able to declare its own status (think by instance about cases where the package could be included from a [VCS](https://getcomposer.org/doc/05-repositories.md#vcs) or a [private repository](https://getcomposer.org/doc/05-repositories.md#loading-a-package-from-a-vcs-repository)).

VincentLanglet:
> I guess we could also add `"abandoned": true` in `composer.json`.

This should be enough, thanks to https://github.com/sonata-project/dev-kit/pull/792

PR #619, opened 2020-09-20
Update keywords

OskarStark:


VincentLanglet:
Imho, we should remove `bootstrap` which is not related.

And we should add `bundle` and `symfony`.

VincentLanglet:
Similar to https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1119

I would propose:
- admin
- admin-generator
- bundle
- doctrine
- phpcr (and odm ?)
- sonata
- symfony

composer.json
```diff
@@ -3,10 +3,11 @@
     "type": "symfony-bundle",
     "description": "Symfony Sonata / Integrate Doctrine PHPCR into the SonataAdminBundle",
     "keywords": [
-        "Admin Generator",
         "admin",
+        "admin-generator",
+        "bootstrap"
```
jordisala1991, line 8: the syntax is broken, missing commas

PR #641, opened 2020-11-13
DevKit updates for 2.x branch

SonataCI:


VincentLanglet:
@dbu Do you know how to fix these tests ?

I saw you doing multiple PR about this issue.

dbu:
hm, i have not looked into this in a while... it seems to me like during the migration from travis-ci to github actions, part of the setup steps has been lost. the ci build needs to run some setup steps: https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/blob/2.x/.travis/before_script_test.local.sh#L9-L10

i guess we should add those commands into the makefile? do you want to give it a try?

VincentLanglet:
> hm, i have not looked into this in a while... it seems to me like during the migration from travis-ci to github actions, part of the setup steps has been lost. the ci build needs to run some setup steps: https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/blob/2.x/.travis/before_script_test.local.sh#L9-L10
> 
> i guess we should add those commands into the makefile? do you want to give it a try?

Thanks !

dbu:
great! 
the PHP 8 build should not even be here, as we don't support php 8 in composer.json (and phpcr-odm also does not yet support it anyways). can you remove that build? i think everything else looks fine!

VincentLanglet:
> great!
> the PHP 8 build should not even be here, as we don't support php 8 in composer.json (and phpcr-odm also does not yet support it anyways). can you remove that build? i think everything else looks fine!

This is not a required build. It's here for every sonata bundle in order to know if the bundle is compatible with php 8.

dbu:
can we mark it as allowed to fail, like with travis-ci? currently it marks the PR as "Merging is blocked" and shows in red font as if it was a problem, when actually it is more of an information.

VincentLanglet (dismissed):

VincentLanglet:
> can we mark it as allowed to fail, like with travis-ci? currently it marks the PR as "Merging is blocked" and shows in red font as if it was a problem, when actually it is more of an information.

Merged was blocked because it needs a review.

For the red fonts, I don't know.

VincentLanglet:
https://github.com/sonata-project/dev-kit/pull/1253 will be needed first

VincentLanglet (approved):

PR #683, opened 2021-03-17
DevKit updates for 2.x branch

SonataCI:


phpunit.xml.dist
```diff
@@ -22,7 +28,6 @@
 
     <php>
         <ini name="precision" value="8" />
-        <env name="KERNEL_CLASS" value="Sonata\DoctrinePHPCRAdminBundle\Tests\Fixtures\App\Kernel" />
```
jordisala1991, line 25: Either this env var is defined in custom bootstrap somehow , added to dev-kit or this file cannot be generic

VincentLanglet: I tried 
```
public static function getKernelClass(): string
    {
        return Kernel::class;
    }
```

Seems like it doesn't work...

VincentLanglet: @sonata-project/contributors Any idea ?

OskarStark (dismissed):

VincentLanglet (changes requested):
Should not be merged like this

VincentLanglet (approved):
Finally

jordisala1991 (approved):

